PR #3, opened 2020-08-31
fix(web): adds support for preview to templates

t0ff3r:
We need to be able to preview content in contentful before it is build by CI

Preview works for articles, forms, categories, main categories and frontpages

## How is it solved?
By adding a secret path /preview/:entryId to our app where we can load pages directly from contentful using the preview token. To use this feature a new environmentvariable called CONTENTFUL_PREVIEW_TOKEN needs to be set in .env. Already set this up in test and prod

| Environment variables               |   x    |

- [x] Pick an ID from one of the entries in contentful and go to the correlating domain from the azure pipelines bot and add /preview/your-entry-id to the end of the url. You can then edit content in contentful and refresh your preview to see the changes.


OysteinThuen:
Build #3037 successfully built and deployed.

- Prod: https://5f4e1ccf49ba2ff981857437--elvia-prod.netlify.app
- Test: https://5f4e1c85fd2dd0e5fe2402a0--elvia-test.netlify.app

packages/web/gatsby-node.js
```diff
@@ -541,12 +532,20 @@ exports.sourceNodes = async ({
...
 
-  await getMainCategories().then(createMainCategories)
-  await getCategories().then(createCategories)
-  await getArticles().then(createArticles)
-  await getFronts().then(createFronts)
-  await getForms().then(createForms)
-  await getTerms().then(createTerms)
-  await getCustomerSegments().then(createSegments)
-  await getMenus().then(createMenus)
+  const transformer = createTransformer({
```
JonatanPaalsson, line 535: You could get these values from `config/contentful.ts`. Up to you 😄

packages/web/contentful/fetch.js
```diff
@@ -33,14 +24,27 @@ const reportFetchErrors = (name, errorList) => {
   }
 }
 
-const fetchEntriesOfType = async type => {
-  const config = createConfig()
-  const client = contentful.createClient(config)
-  const entries = await client.getEntries({
-    content_type: type, // eslint-disable-line
+const fetchEntriesOfType = curry(async (client, type, id) => {
+  let options = {
```
JonatanPaalsson, line 28: Why let and not const?

JonatanPaalsson (changes requested):

JonatanPaalsson:
I only get redirected to the frontpage when following the steps in How it can be tested and verified 😕

t0ff3r:
> I only get redirected to the frontpage when following the steps in How it can be tested and verified 😕

Just realised the same thing. Looking into it

packages/web/gatsby-node.js
t0ff3r: I know but this part is about to be broken out to its own package. @FredricW will use the correct config when he sets it up the new way

packages/web/contentful/fetch.js
FredricW: let it be



;)

JonatanPaalsson (approved):

JonatanPaalsson (changes requested):
Mangler fallback om ID ikke finnes

FredricW:
To ting: Er det ikke litt forvirrende å bli tatt til forsiden hvis id ikke finnes? Jeg trodde jeg hadde skrevet feil url flere ganger.
Men hvis det likevel er ønskelig, skal det også være fallback på `/preview` (som for øyeblikket er hvit)?

Hadde det vært enklere å forstå om den sa at id ikke finnes?

FredricW (approved):
LGTM. Kan eventuelt fikse fallback greiene i en senere pr

packages/web/contentful/fetch.js
```diff
@@ -33,14 +24,27 @@ const reportFetchErrors = (name, errorList) => {
...
     include: 5, // 4 is needed for resolving aliased data on shortcuts in sections
     limit: 1000,
-  })
+  }
+
+  if (type) {
+    options = {
```
FredricW, line 34: Isn't mutating existing variables like this kind of an antipattern?

FredricW: I mean, it is not straight forward to know what the options object will look like when all of this goes through

PR #10, opened 2020-09-01
Refactor/omega 74 make new layouts

lobunto:
OMEGA-74

To separate Layout components for _public_ content and _Min side_ content.

- There should be a PublicLayout component and MinSide component which wraps the different part of the website.
- `minSideNavigation` prop passed to MainHeader in both Layouts are hardcoded as of now and will be fixed in a separate PR for split of Header.

## How is it solved?
- Moved Min side specific content to a `MinSideLayout` component
- Moved public specific content to a `PublicLayout` component
- Both `MinSideLayout` and `PublicLayout` are both exported from `components/Layout/index.ts` so it is easier to related to them

| Unit tests                          |     |

- [x] Everything for the public pages should look like before - no style changes
- [x] Everything for the min side pages should look like before - no style changes


JonatanPaalsson (approved):
LGTM!

PR #18, opened 2020-09-02
fix(web/redirect.js): fix redirect from /kart/ to strombruddskart

lobunto:
OMEGA-X

Missing redirect from `/kart/` to `/strombruddskart`

`/kart/` should redirect to `/strombruddskart`

## How is it solved?
Add redirect from `/kart/` to `/strombruddskart`

- [ ] path `/kart/` should redirect to `/strombruddskart`


lobunto:
Fix not needed. Seems there where some cache issues for the redirects just after release.

PR #19, opened 2020-09-02
open links in form description in new tab

JonatanPaalsson:
Data in schema is lost when opening description link

Should open links in form description in a new tab

## How is it solved?
passing in openInNewTab prop to renderRichText in Form template

- [ ] compare with https://www.elvia.no/nettleie/alt-om-nettleie/krav-om-redusert-satsfritak-fra-avgift-pa-elektrisk-kraft


lobunto (changes requested):
Probably this logic should live inside the Link component? All external links should open in a new tab or what?

JonatanPaalsson:
> Probably this logic should live inside the Link component? All external links should open in a new tab or what?

I think only the Links in form should open in a new tab. This way our customers don't need to refill the form again if they navigate to another site

lobunto (approved):

PR #23, opened 2020-09-03
refactor(contentful-transformer): moved transformer to its own package

FredricW:
OMEGA-72

There are multiple reasons that the transformer benefits from being its own package:

* Separation of concern - It brings the data from CMS into the app during build
* Uses a different language (javascript instead of typescript) - which means it is not interoperable with the rest of the project
* Lower complexity of the web setup file (`gatsby-node.js`)
* Gatsby plugin system - Abstracts the details from the web app that is not important to have there

1. Repo should have a new package called `contentful-source`
2. The package should be able to be used as a gatsby plugin
3. The package should be able to be used directly as a regular npm package

## How is it solved?

* A new package: `contentful-source`, now exists in the repo
* It is now used as a gatsby plugin (see `gatsby-config.js`) and the environment details are passed in to the plugin options object

| Dependencies          |  x       |
| Environment variables | x        |

- [ ] Project should build as normal (may need to run `yarn` first): `yarn web:clean && yarn web:develop`
- [ ] `/preview/[some-id]` should still work as expected (may need to add `CONTENTFUL_PREVIEW_TOKEN` environment variable)


t0ff3r (approved):
Lgtm

PR #5, opened 2020-09-01
Feature/omega 27 customer relation overview page

ragnhildlarsen:
OMEGA-27

## What is the reason for this change?
Customers that have multiple customer relations need an overview page for all their customer relations.  

### What is the expected result?
Customers with multiple customer relations should se an overview page if going to /minside/kundeforhold. Customers with only one customer relation should be redirected to the only customer relation they have /minside/kundeforhold/customerId. Customers with no customer relations should be redirected to access request page /minside/accessrequest. When you click on a custom relation card on the overview page you should be navigated to a detail page with an url /minside/kundeforhold/customerId which currently only shows "test".

[Figma sketches](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=9334%3A57402)

## How is it solved?
A new template is created for CustomerRelationOverview. This component accesses the customer store and uses that data to decide what should be shown or if the customer should be redirected. 
A dummy template for CustomerDetail have been created to be a placeholder for the detail page for a Customer Relation (this might be excessive if someone else is working on the actual page). 

## Any changes/updates that may affect others?
|                                     |Changed|
|-------------------------------------|-------|
| Unit tests                          |      |
| Dependencies                        |       |
| Environment variables               |       |
| Build scripts                       |       |

## How can it be tested and verified?

- [ ] Test behaviour for different test customers and check that the correct data is shown.
- [ ] UI looks and behaves identical to design defined in Figma.
- [ ] enter /minside/kundeforhold in the url to navigate to the page
- [ ] test user Per Hansen with birth number 05039703930 has 1 private and 1 organization customer relation
- [ ] test user Ole UtenKontrakt with birth number 23028791364 has no customer relations
- [ ] test user Testprivat122 with birth number 23025139593 has 1 customer relation
- [ ] functionality is only visible in development environment

PR #26, opened 2020-09-03
OMEGA-89: New layout for Min Side

eTallang:
OMEGA-89

A new layout is needed for the new version of min side.

The implementation should look like the designs in [Figma](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=9560%3A55962).

- [ ] The implementation complies with the designs in Figma.
- [ ] URL's in the breadcrumbs and tabs makes sense
- [ ] Active tab is always visible and correct, even when navigating through the URL
- [ ] Should the tabs be made into a new component? If so, where and called what (reminding that it already exists a Tab component in hdd-react).
- [ ] All elements are accessible via keyboard.


eTallang:
**NOTE: ** Merge after [OMEGA-27](https://github.com/3lvia/kunde-elvia-web/pull/5), since this PR contains code from that branch.

packages/web/src/pages/minside.tsx
```diff
@@ -60,6 +62,22 @@ const Minside: React.FC<PageContext> = props => {
...
+              component={CustomerDetail}
+            >
+              {/* TODO: Add child routes to min side here */}
+              <PrivateRoute path="/" component={Meterpoint} />
+              <PrivateRoute path="/malere" component={Meterpoint} />
+            </PrivateRoute>
+          </>
+        )}
         <PrivateRoute path="/minside/meterpoint" component={Meterpoint} />
         <PrivateRoute path="/minside/hanport/:contractId" component={HANPort} />
```
lobunto, line 82: We should probably find out if we want to use norwegian or english in the url

packages/web/src/templates/CustomerDetail/CustomerDetail.module.scss
```diff
@@ -0,0 +1,83 @@
...
+  margin-bottom: spaceCalc(12);
+}
+
+.tabContainer {
+  white-space: nowrap;
+  overflow-x: auto;
+  margin-bottom: spaceCalc(9.5);
+}
+
+.tab {
```
lobunto, line 27: I think the tabs, according to the design, have some padding (8px/`spaceCalc(1)`) on the sides..?

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -0,0 +1,126 @@
...
+    },
+    {
+      name: 'Tilganger',
+      url: '/tilganger',
+    },
+    {
+      name: 'HAN-port',
+      url: '/han-port',
+    },
+  ]
```
lobunto, line 59: As mentioned about the language above, we need some consistency for the urls. Above we have `/minside/hanport` here we have `/han-port`. Maybe we should use what is defined in the `basePaths.json`?

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -0,0 +1,126 @@
...
+          [css.hidden]: !isTablet && breadcrumbs.length === 1,
+        })}
+      >
+        {breadcrumbs.map((breadcrumb, index) => {
+          return (
+            <Link to={breadcrumb.url} key={index}>
+              {breadcrumb.name}
+            </Link>
+          )
+        })}
```
lobunto, line 96:
```suggestion
        {breadcrumbs.map((breadcrumb, index) => (
            <Link to={breadcrumb.url} key={index}>
              {breadcrumb.name}
            </Link>
          )
        )}
```

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  display: grid;
+  gap: spaceCalc(2);
+  grid-template-columns: 1fr;
+  margin: 0 auto spaceCalc(7.5) auto;
+  max-width: emCalc(400px);
+  @media #{$tablet} {
+    margin: 0 0 spaceCalc(7.5);
+    max-width: none;
+    gap: spaceCalc(4);
+    grid-template-columns: 1fr 1fr 1fr 1fr;
```
lobunto, line 37: Is this going to repeat for several rows? Then it could be written `grid-template-columns: repeat(4, 1fr)`

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  }
+  @media #{$desktop} {
+    gap: spaceCalc(7);
+    grid-template-columns: 1fr 1fr 1fr;
```
lobunto, line 41: Same here as above (if that is the case)

lobunto (changes requested):

packages/web/src/templates/CustomerDetail/CustomerDetail.module.scss
eTallang: Good catch!

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.module.scss
eTallang: Should be fixed through [OMEGA-27](https://github.com/3lvia/kunde-elvia-web/pull/5)

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.module.scss
eTallang: Should be fixed through [OMEGA-27](https://github.com/3lvia/kunde-elvia-web/pull/5)

packages/web/src/pages/minside.tsx
eTallang: I guess we landed on Norwegian. Do you think we should do any further changes to this PR then?

lobunto: Seems so. Then we need to add some redirects as well, because customers may have SMS's containing especially links to `/minside/accessrequest`, maybe this could redirect to `/minside/tilgangstyring`.

lobunto (changes requested):
Routes for `/minside/invoices` should also be renamed to norwegian.

lobunto (approved):

PR #31, opened 2020-09-04
Refactor/omega 26 mypage header redesign cleanup

lobunto:
OMEGA-26

Part 2 of header redesign related to OMEGA-26. 

Min side header should have the same look like the sketches in [Figma](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=10181%3A68249) 

## How is it solved?
- Remove menu from the Min side header.
- Fix styling according to the design
- Remove deprecated css classes
- Refactor "local" components and use them directly in render return instead.

- Nope

- [ ] Min side header should have the same look like the sketches in [Figma](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=10181%3A68249)


t0ff3r (approved):
Deleting code ❤️

PR #33, opened 2020-09-04
Fix/omega 70 restyle list

lobunto:
OMEGA-70

List should use correct styles. Instead of using Elvis directly, the List css has been copy/paste into our project.

- OL lists should have correct alignment with dots
- UL lists should have correct alignment with numbers

## How is it solved?
Copied Elvis list styles and made them work in our project

Nope

Go to this URL for testing: /proff/planlegge-prosjekter/planlegging/slik-jobber-elvia-–-en-oversikt-over-prosessen

- [ ] OL lists should have correct alignment with dots
- [ ] UL lists should have correct alignment with numbers


t0ff3r (approved):
Looks great! Seems like removing need for Elvis saved us a lot of lines of code 🙈

PR #51, opened 2020-09-08
fix: removes deps from web and adds contentful to contentful package

t0ff3r:

Clean up some dependencies in web

Web should not install packages not in use

## How is it solved?

Removed gatsby-source-contentful, contentful-management, contentful and prop-types. Added contentful to contentful-source package

| Dependencies          |     x    |

- [ ] Hopefully everything builds as normal


packages/web/package.json
```diff
@@ -48,21 +46,17 @@
...
     "gatsby-plugin-typescript": "^2.4.3",
-    "gatsby-source-contentful": "^2.3.12",
     "gatsby-source-filesystem": "^2.3.8",
     "gatsby-source-graphql": "^2.5.2",
     "gatsby-transformer-remark": "^2.8.13",
     "graphql": "^14.5.8",
     "identity-obj-proxy": "^3.0.0",
     "normalize.css": "^8.0.1",
     "oidc-client": "^1.10.1",
-    "prop-types": "^15.7.2",
```
JonatanPaalsson, line 63: We are using prop-types in some components in hdd-react, should we then move this dependency to hdd-react? Bit unsure how this actually works

packages/web/package.json
```diff
@@ -48,21 +46,17 @@
     "gatsby-plugin-netlify": "^2.3.3",
     "gatsby-plugin-prefetch-google-fonts": "^1.4.3",
     "gatsby-plugin-react-helmet": "^3.3.2",
-    "gatsby-plugin-sass": "^2.3.2",
```
JonatanPaalsson, line 51: we don't need this? 🤯

JonatanPaalsson (changes requested):

packages/web/package.json
t0ff3r: Yes - lets move it.

JonatanPaalsson (approved):

PR #52, opened 2020-09-09
OMEGA-97 Manual reading same value as last time

erlbe:
OMEGA-X

Request for being able to submit same manual reading value as last time.

image

Allowed to submit same value as before.

image

## How is it solved?

Changed validation. Both frontend and backend. (Backend in production)


JonatanPaalsson (approved):

JonatanPaalsson:
LGTM! Remember to update the term in contentful after merge 😄

PR #60, opened 2020-09-10
feature/OMEGA-99 add emptyInvoiceBox component

marasva:
[OMEGA-99](https://elvia.atlassian.net/browse/OMEGA-99?atlOrigin=eyJpIjoiMTJmMTgwYWE4ODc0NDMxYzkwNTUwNWY2ZjVkOWJjM2QiLCJwIjoiaiJ9)

Customers that has not yet received an invoice is shown an empty table. 
image

A new view with text and icons explaining that the customer have not yet received an invoice.

image

## How is it solved?

Add new Begrep in contentful and add a new component with icons from Figma.

- [ ] Use the test user Per Hansen and go into his meterpoint view for Ole Nilsens gate 21 (222222222)
- [ ] A new view will show with the same design as above
- [ ] Looks as usual (with a table view of invoices) if you try some of the other meterpoints.


marasva:
I changed the @ media to 1120px because I could not find a way to keep the design without breaking it at the breakpoint between 1024px - 1060px through using the regular max-width for MobileAndTablet.

packages/hdd-react/src/components/EmptyInvoiceBox/EmptyInvoiceBox.tsx
```diff
@@ -0,0 +1,67 @@
...
+  style,
+}) => {
+  useStyles(css)
+  const classes = classnames(css.emptyInvoice, className)
+
+  return (
+    <>
+      <div className={classes} style={style}>
+        <div className={css.leftIconsEmptyInvoice}>
+          <div>
```
eTallang, line 33: Are these `<div>` elements that wrap the icons necessary? :slightly_smiling_face: Maybe we could do something like:
```
<HddSendEmailLargeColor />
<HddSendEmailSmallColor className={css.sendEmailLeft2} />
<HddPaperPlaneOrangeSmallColor className={css.paperPlaneLeft} />
```

marasva: I cannot manage to make it work without them, but I agree that it looks unnecessary

packages/ui/src/components/EmptyInvoiceBox/EmptyInvoiceBox.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const EmptyInvoiceBox: React.FC<EmptyInvoiceBoxProps> = ({
+  className,
+  style,
+  infoText,
+}) => {
+  const classes = classnames(css.emptyInvoice, className)
+
+  return (
+    <>
+      <div className={classes} style={style}>
```
JonatanPaalsson, line 30: Is it possible to remove some of these divs? Some icons are just wrapped in a empty div and maybe the classnames can go directly on the icons?

JonatanPaalsson (changes requested):

lobunto:
@marasva what is the status for this PR?

marasva:
Well, I do not know of any "smoother" way to implement the svgs without so many divs, so for me it is finished. Do not think @JonatanPaalsson have looked further on it either?

t0ff3r:
Seems overly complex to import and place every icon freely one by one. I am no expert on svgs but could we solve this by making a nested svg instead of handling each of them separately? 

This article seems interesting: https://www.sarasoueidan.com/blog/mimic-relative-positioning-in-svg/

marasva:
> 
> 
> Seems overly complex to import and place every icon freely one by one. I am no expert on svgs but could we solve this by making a nested svg instead of handling each of them separately?
> 
> This article seems interesting: https://www.sarasoueidan.com/blog/mimic-relative-positioning-in-svg/

Will look into the article you sent.

JonatanPaalsson (approved):
LGTM!

PR #66, opened 2020-09-11
added new pilotgraph component

JonatanPaalsson:
OMEGA-130

We're working on the pilot project for mypage

New component which is shown in the gif

pilotgraf

## How is it solved?

This component is basically a copy of the ConsumerGraph component but it hooks into the drillUp- and drilldown events from highcharts. On drilldown I check if we are on the lowest level then add all plotBands and if we drillUp then I remove all of them.

Exactly where the data about the plotbands should exist is unclear at the time  and needs to be extended anyways, therefor I just added them directly in the component for now if anyone wants to test it out.

- [ ] Meterpoint on mypage should work as before


lobunto (approved):
LGTM

PR #67, opened 2020-09-11
feat(web): add table with meters

eTallang:
OMEGA-33

A new table with meterpoints is required for the restructure of Min Side.

A new table should be visible with a list of all meterpoints.

- [ ] Check that the implementet table looks and behaves as designed in [Figma](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=9560%3A55962).
- [ ] Check that the table is keyboard accessible.
- [ ] Check that clicking a row directs to the correct meterpoint page.
- [ ] Check that the correct meterpoints are visible in the table.
- [ ] Check that the table is responsive according to the designs.


packages/web/src/templates/Meterpoints/Meterpoints.module.scss
```diff
@@ -0,0 +1,53 @@
+@import '../../theme';
+@import '@elvia/elvis/src/utilities/breakpoints.scss';
+
+.tableContainer,
+.table {
```
lobunto, line 5: Is this table styling special for the meteringpoint list?

packages/web/src/templates/Meterpoints/Meterpoints.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (event.key === 'Enter') {
+      goToMeterpoint(id)
+    }
+  }
+
+  const getAddress = (meteringPoint: MeteringPoint) => {
+    const address = meteringPoint.address
+    return `${address.streetName} ${address.buildingNumber}${address.letter ??
+      ''}`
+  }
```
lobunto, line 37: This is great! 👍

packages/web/src/templates/Meterpoints/Meterpoints.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const onKeyDown = (
+    event: React.KeyboardEvent<HTMLTableRowElement>,
+    id: string
+  ) => {
+    if (event.key === 'Enter') {
+      goToMeterpoint(id)
+    }
+  }
```
lobunto, line 31: This is great! 👍

packages/web/src/pages/minside.tsx
```diff
@@ -87,8 +88,8 @@ const Minside: React.FC<PageContext> = props => {
           path="/minside/kundeforhold/:customerId"
           component={CustomerDetail}
         >
-          <Redirect path="/" from="/" to="malerpunkt" noThrow />
-          <PrivateRoute path="malerpunkt" component={Meterpoint} />
+          <Redirect path="/" from="/" to="malere" noThrow />
+          <PrivateRoute path="malere" component={Meterpoints} />
```
lobunto, line 92: Will this route be default even if there is only one facility?

lobunto (commented):
LGTM. Got a question tho about the route for meteringpoints

packages/web/src/pages/minside.tsx
eTallang: Thats a good question. I haven't heard anything else, but I'll check with Malin.

eTallang: The designers have discussed it, and if we are to have an exception for customers with one meteringpoint, it's gonna result in three different states for the min side landing page. So lets take a closer look at this later :slightly_smiling_face:

ragnhildlarsen:
I think you need to ensure that the meters shown is the one connected to the customer relation that you are looking at. 

E.g. Per Hansen only have 2 meterpoints himself, and the rest in the list are related to the other customer relationships. 

Screenshot 2020-09-11 at 15 28 06

eTallang:
> I think you need to ensure that the meters shown are connected to the customer relation that you are looking at.
> 
> E.g. Per Hansen only have 2 meterpoints himself, and the rest in the list are related to the other customer relationships.
> 
> Screenshot 2020-09-11 at 15 28 06

A good catch! Fixing it now.

lobunto (approved):

PR #68, opened 2020-09-11
Feature/omega 119 poweroutageapp package

FredricW:
OMEGA-119

We need a better poweroutage map. This is the fundament for future work on the new poweroutage map.

Poweroutage map should render some text instead of the old map if the environment variable `POWEROUTAGE_V2` is set to `true`.

## How is it solved?

Adds a new package in the project: `poweroutage`.
It exposes the `PowerOutage`component which takes a dictionary of terms on the `terms` prop.

| Dependencies          | x       |
| Environment variables | x       |

- [ ] Power outage map should show some text instead of the old map app when environment variable `POWEROUTAGE_V2` is set to `true`
- [ ] The `PowerOutage` component should accept terms from the web and be able to render them.


packages/web/src/utils/richTextRender.tsx
```diff
@@ -272,8 +273,18 @@ export const renderRichText = (
           case ComponentTypes.reactComponent: {
             const { type } = node.data
 
+            console.log('reactComponent: ', type, PowerOutage)
```
t0ff3r, line 276: This is a console log

packages/poweroutage/src/components/yolo/index.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { useContext } from 'react'
+import { TermsContext } from '../../'
+
+export const Yolo = () => {
+  const { title, subTitle } = useContext(TermsContext)
+  return (
+    <h1>
+      Yolo says: {title} and {subTitle}
```
t0ff3r, line 8: What the hell is Yolo?

FredricW: It is the solution to all our problems.

t0ff3r (approved):

PR #72, opened 2020-09-14
Fix table style

steinmy:
Some fixes to the style of the table component
Note: The css in the table component stylesheet did not pass the style linter. This was due to code already present, not due to the changes introduced in this pull request. After discussion with other frontenders, I have disabled the style-lint rule "no-descending-specificity". Feel free to comment if you disagree.

OMEGA-138

Table component is not as it should according to sketcehs
- No borders between rows on mobile view
- On mobile view the first row is only left aligned when containing a `<p>` inside a `<div>`
- Header has too low height
- The first column expands to take all extra horizontal space, instead of distributing between columns

image
image

Tables should look the same as in the Figma sketches.

https://www.figma.com/file/gh5MaG2NWKVODTk1p3cO6s/Tables?node-id=3%3A292

## How is it solved?

Changed stylesheet of table component

image
image

- [ ] Tables at for example /mitt-elvia/prisoversikt look like the screenshots under "How is it solved?" above


.stylelintrc.js
```diff
@@ -23,5 +23,6 @@ module.exports = {
         ],
       },
     ],
+    'no-descending-specificity': null,
```
eTallang, line 26: How come we have disabled this lint-rule? :slightly_smiling_face:

packages/hdd-react/src/components/Table/Table.module.scss
```diff
@@ -78,6 +77,10 @@ $table-border-radius: 8px;
 
   td {
     text-align: right;
+    border-top: 1px solid map-get($colors, 'grey05');
+    &:first-child {
+      text-align: left;
+    }
```
eTallang, line 83: Maybe we could concat this with the CSS on line 87?

```
...
  &:first-child,
  &:first-child div p {
    text-align: left;
  }
```

Or maybe your contribution even solves both scenarios, and the rule on line 87 can be removed entirely?

eTallang (changes requested):

.stylelintrc.js
steinmy: Let's discuss this in person :)

packages/hdd-react/src/components/Table/Table.module.scss
steinmy: I tried removing line 87 to see if this new code is enough, but it didn't seem to work, as the `<p>` has an inline style telling it to align to the right. I'm not sure where this comes from, probably from Contentful somehow?

eTallang (approved):

PR #76, opened 2020-09-15
enable pilotgraph for pilotusers

JonatanPaalsson:

Pilotusers should see the new pilotgraph

No changes in production since there are no pilot users yet. Users should get different versions of the pilotgraph depending on their tariffgroup

## How is it solved?

Check in redux store if the contract is a pilotcontract


lobunto (approved):
Seems none of the "Aktive Hjem" test users does have any consumption data so I cannot see if the graph shows up. Otherwise, it looks good.

PR #80, opened 2020-09-15
Feature/omega 120 power outage map v2

FredricW:
OMEGA-120

Needs map component in poweroutage.

Power outage map v2 should now show a simple map.

## How is it solved?

Uses `GoogleMap` component from `hdd-react`. Updated existing default center to point to Oslo instead of somewhere in the US.

- [ ] It should now show a simple map: 
image

- [ ] The map should be centered at Oslo.


packages/hdd-react/src/components/GoogleMap/GoogleMap.tsx
```diff
@@ -27,8 +27,8 @@ export const GoogleMap: React.FC<GoogleMapProps> = ({
     language: 'nb-NO',
   },
   defaultCenter = {
-    lat: 59.95,
-    lng: 30.33,
+    lat: 59.918,
+    lng: 10.734,
```
t0ff3r, line 31: This could maybe be configurable in the cms in the future? 🤔

t0ff3r (approved):

packages/hdd-react/src/components/GoogleMap/GoogleMap.tsx
FredricW: Yes, good point

PR #83, opened 2020-09-15
Feature/omega 150 info component power outage

corneliusgdahling:
OMEGA-150

Create an infobox for the new power outage map

image
image

Should have info box

## How is it solved?

Created info box

**NB: Table is added as children and is not styled as it is supposed to be, but instead is an example. Width of box is decided by parent**

image
image

View in new poweroutage


packages/poweroutage/src/components/PowerOutageInfo/PowerOutageInfo.module.scss
```diff
@@ -0,0 +1,64 @@
...
+  display: flex;
+  flex-direction: column;
+  background-color: map-get($colors, 'white');
+
+  .titleContainer {
+    background: map-get($colors, 'redSignal');
+    border-radius: 8px;
+    display: flex;
+
+    &-Maintenance {
```
erlbe, line 14: Should this be named "Planned" instead of Maintenance? To correspond with the type names

corneliusgdahling: Fixed

packages/poweroutage/src/components/InfoBox/InfoBox.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { FC, useContext } from 'react'
+import classnames from 'classnames'
+import dayjs from 'dayjs'
+import * as css from './InfoBox.module.scss'
+import { TermsContext } from '../../'
+import { TitleSmall, TextInfo } from '@elvia/ui/src/components/Typography'
+import { ElvisIcon } from '@elvia/ui/src/components/ElvisIcon'
+import { Button, ButtonType } from '@elvia/ui/src/components/Button'
+
+export enum OutageType {
```
erlbe, line 10: Once this PR (https://github.com/3lvia/kunde-elvia-web/pull/87) is merged, we can use the type `types.ts` . So we don't need separate enums in each component.

erlbe: Unless we actually want to keep separate types in components :)

corneliusgdahling: Fixed

erlbe (approved):

PR #89, opened 2020-09-16
fix(cli): environment path is now relative and platform independent

FredricW:
OMEGA-X

CLI isn't able to find environment variables.

CLI should find environment variables and not crash.

## How is it solved?

Config path is now relative to the config file and uses `path` for cross platform support.

| Build scripts         |  x       |

- [ ] Running the CLI should not throw any errors (except `hdd create` on windows which is another issue).


corneliusgdahling (commented):
Should we also add a fix here, so it does not complain about access token, when it has been renamed to management token?
image

t0ff3r (approved):

PR #90, opened 2020-09-16
Feat/omega 146 meteringpoint table filter

eTallang:
OMEGA-146

Previously, there was no way to filter the list of metering points in the metering points table. This is a desired functionality, especially for clients with many metering points.

The metering points table should now have a filter above.

image

- [ ] Check that the implementation matches the design in [Figma](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=10181%3A67879)
- [ ] Check that the input can filter the list on all specified props (metering point no, address and metering point id)
- [ ] Check that the implementation follows best practices
- [ ] Check that the implementation is responsive
- [ ] Check that the implementation is performant.


packages/web/src/templates/Meterpoints/MeteringPoints.tsx
```diff
@@ -16,36 +17,79 @@ interface MeteringPointsProps {
...
+      <div className={css.tableContainer}>
+        <Table className={css.table}>
+          <thead>
+            <tr>
+              <th>Målernummer</th>
+              <th>Adresse</th>
+              <th>Post nr.</th>
+              <th>Målepunkt-ID</th>
+              <th>Forbruk hittil i {dayjs().format('MMM')}</th>
+              <th className={css.actionColumn}></th>
```
t0ff3r, line 77: Shouldn't this be terms from contentful?

lobunto:
@eTallang do you know of that we do have a useSearch hook? Could we probably consider using that here as well..

packages/web/src/templates/Meterpoints/Meterpoints.module.scss
```diff
@@ -1,6 +1,15 @@
...
 @import '@elvia/elvis/src/utilities/breakpoints.scss';
 
+.filter {
+  max-width: spaceCalc(72);
+  margin-bottom: spaceCalc(8);
+}
+
+.filterIcon {
+  margin-top: spaceCalc(0.5) * -1;
+}
```
ragnhildlarsen, line 11: Looks like the search/filter icon should have more padding-right (16px/24px) on desktop (it's a bit unclear in the sketches so maybe double check with designer).

ragnhildlarsen:
Should there be any empty state when a search give no result. Maybe outside the scope of this task but I was just thinking about it :) 
Screenshot 2020-09-17 at 08 48 29

lobunto:
@eTallang are you following up the comments here..? :)

packages/web/src/templates/Meterpoints/Meterpoints.module.scss
eTallang: Yes I noticed this during development, however input-field is a component with a icon-prop, so I wanted to stay consistent with the other places we use icons in a text field. I'll check this with designer.

packages/web/src/templates/Meterpoints/MeteringPoints.tsx
```diff
@@ -16,34 +18,80 @@ interface MeteringPointsProps {
...
+  const { result: filteredContracts, search } = useSearch({
+    data: contracts,
+    bypassEmptyQuery: true,
+    options: {
+      keys: [
+        'meteringPoint.meterSerialNumber',
+        'meteringPoint.meteringPointId',
+        'meteringPoint.address.streetName',
+        'meteringPoint.address.buildingNumber',
+        'meteringPoint.address.letter',
```
ragnhildlarsen, line 46: Should it be possible to search on zipCode?

ragnhildlarsen:
Looks like placeholder text should be truncated on mobile: [https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=9755%3A61892](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=9755%3A61892)
Screenshot 2020-09-23 at 10 05 40

ragnhildlarsen (changes requested):
Looks good! Just had a few comments about placeholder text on mobile and being able to search on zip code. Code looks good to me.

ragnhildlarsen (approved):
Looks good to me :)

lobunto:
Should the search field be visible for even users with a single contract? A suggestion is to show it for those with 5 or more ...

eTallang:
> Should the search field be visible for even users with a single contract? A suggestion is to show it for those with 5 or more ...

I think the plan is that when you have few contracts, you won't see the table at all (replace the tab with a tab to see a single meters consumption). But maybe we should confirm this with design.

eTallang:
We've decided to hide the filter until the task for replacing the table with the metering point page has been completed. This is because the filter does not serve any purpose for most of Elvias customers, since the majority only have one metering point.
This PR does not use the new flag-feature, but uses a poor-mans flag (a `if (false)` statement). @t0ff3r would you still take a look to get this PR ready for merge?

lobunto (approved):

packages/ui/src/components/FormInputs/FormInputs.module.scss
```diff
@@ -23,9 +23,9 @@ fieldset {
   font-size: emCalc(16px);
   height: emCalc(48px);
   outline: none;
-  margin: spaceCalc(0.5) 0 spaceCalc(1) 0;
   padding: emCalc(12px) emCalc(16px);
   width: 100%;
+  text-overflow: ellipsis;
```
lobunto, line 28: Discovered just now that this makes the prefix input hide the +47.. :p 

Screenshot 2020-09-25 at 15 03 46

lobunto (changes requested):

packages/ui/src/components/FormInputs/FormInputs.module.scss
lobunto: So it will probably also affect every form that uses the phonenumber component

eTallang: Absolutely. A terrifying side effect I should have anticipated. We will have to target the inputs that only have a placeholder.

eTallang: I've fixed it by slightly increasing the width of the input.

packages/web/src/hooks/useSearch.tsx
```diff
@@ -100,39 +115,52 @@ export const fuzzySearch = ({
...
+      search: setTerm,
+      term,
+      reset,
+    }
+  } else {
+    const filteredData = locale
+      ? data?.filter?.(
+          dataItem =>
+            (dataItem as any).locale === locale ||
+            (dataItem as any).locale === null
```
t0ff3r, line 147: why are we casting this to any?

eTallang: A cast to `any` is "necessary", since `dataItem` is of type `T`. A better type cast would be to say`dataItem as SearchDataItem`, however type T and SearchDataItem has no overlap, so the IDE complains. To fix this, I could convert it to `unknown` first, but then it turns out to be a very long line of code.

Considering that the `SearchDataItem` is an interface that doesn't provide any big typing support, a cast to `any` made sense to me instead of going with the first approach.

packages/web/src/templates/Meterpoints/MeteringPoints.tsx
```diff
@@ -16,34 +18,79 @@ interface MeteringPointsProps {
...
-            accessToken={accessToken}
-            contract={contract}
-            customerId={customerId}
-            key={contract.id}
-          ></MeteringPointRow>
-        ))}
-      </tbody>
-    </Table>
+    <>
+      {false && (
```
t0ff3r, line 56: false && ??

t0ff3r (approved):

PR #91, opened 2020-09-16
fix create script in cli for windows

JonatanPaalsson:

create script does not work for windows

should be able to use `yarn ui:create` on windows

## How is it solved?

Use process.cwd() instead of process.argv[1]

- [x] Should be able to use yarn ui:create on windows
- [x] Should be able to use yarn ui:create on Mac


FredricW (approved):
Works fine on the Mac!

lobunto:
@JonatanPaalsson does this works on Windows now? Probably @eTallang could test as well..?

JonatanPaalsson:
Works on my windows machine 👍 But It would be nice if @eTallang could test aswell!

PR #92, opened 2020-09-16
Migrations/omega 165 add poweroutage v2 to contentful

corneliusgdahling:
OMEGA-165

Migration add StrømbruddskartV2 to Contentful

StrømbruddskartV2 should be option in interactiveComponent in Contentful and should render in rich text if selected in Contentful

## How is it solved?

- Created new content model called interactiveComponent (which is same as reactComponent)
- Created field called interactiveComponentType (but has field id and name type and Type, respectively)
- Added support for rendering StrømbruddskartV2 in richTextRenderer and sections.
- Added interactiveComponent to contentful-source (preview)


FredricW:
Kan vi rename denne snart, før den heter reactComponent overalt?

migrations/fields/interactiveComponentType.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const interactiveComponentType: FieldConfig = {
+  id: FieldIds.Type,
+  options: {
+    name: FieldNames.Type,
+    type: 'Symbol',
+    required: true,
+    validations: [
+      {
+        in: ['Strømbruddskart', 'Log inn knapper', 'StrømbruddskartV2'],
```
lobunto, line 12:
```suggestion
        in: ['Strømbruddskart', 'Logg inn knapper', 'StrømbruddskartV2'],
```

lobunto (changes requested):

packages/web/src/components/SectionsWrapper/SectionsWrapper.tsx
```diff
@@ -104,6 +105,27 @@ export const renderSection = (
           return componentToRender(type)
         },
       ],
+      [
+        typeIs(SectionType.InteractiveComponent),
+        data => {
```
FredricW, line 110: I guess we can fix this later, but we have two switch cases that do the same thing. Should probably be a InteractiveComponent picker or something somewhere.

migrations/fields/interactiveComponentType.ts
corneliusgdahling: This would not work, and would require more refactoring. but will fix

packages/web/src/utils/richTextRender.tsx
```diff
@@ -273,15 +274,27 @@ export const renderRichText = (
...
-                equals('Strømbruddskart'),
-                () =>
-                  process.env.POWEROUTAGE_V2 === 'true' ? (
-                    <PowerOutage />
-                  ) : (
-                    <PowerOutageMap />
+                T,
+                data =>
+                  always(
+                    <div>React component support missing for type: {data}</div>
```
FredricW, line 284:
```suggestion
                    <div>Interactive component received an unknown type: {data}</div>
```

packages/web/src/components/SectionsWrapper/SectionsWrapper.tsx
```diff
@@ -104,6 +105,27 @@ export const renderSection = (
...
+
+          const componentToRender = cond([
+            [equals('Strømbruddskart'), always(<PowerOutageMap />)],
+            [equals('StrømbruddskartV2'), always(<PowerOutage />)],
+            [equals('Logg inn knapper'), always(<Login />)],
+            [
+              T,
+              data =>
+                always(
+                  <div>React component support missing for type: {data}</div>
```
FredricW, line 121:
```suggestion
                  <div>Interactive component received unknown type: {data}</div>
```

migrations/fields/interactiveComponentType.ts
corneliusgdahling: never mind. I just noticed, it is not used anywhere

FredricW, line 12:
```suggestion
        in: ['Strømbruddskart', 'Logg inn knapper', 'StrømbruddskartV2'],
```
Typo?

FredricW (changes requested):
Added some comments

packages/web/src/components/SectionsWrapper/SectionsWrapper.tsx
corneliusgdahling: As a utils function? So it is used both in sections wrapper and richtextrenderer?

FredricW: yes

FredricW (approved):

lobunto (approved):

PR #93, opened 2020-09-16
Fix issues with table width and corners

steinmy:
Fixes issues with table width and corners

OMEGA-162

Limiting the width of tables doesn't work, they always take up the width necessary for the content instead of using a scrollbar. Also the contents of the table poke through the rounded corners in some browsers.

image

* Tables in articles, categories and main categories should have a limited width and a horizontal scrollbar
* Table corners should look ok in different browsers

image

## How is it solved?

* Table component is surrounded by two divs
  * Inner container has `overflow-x: auto;` to get scrollbar
  * Outer container has `overflow: hidden;` to prevent anything from poking through the corners
* Table borders are moved to the outer container
* Table component gets new property `isLimitedWidth` which sets the width to what it should be for articles, categories and main categories
* This property is set to true when tables are rendered with `richTextRenderer`

Tables have whole corners and are scrollable if too wide. Some places where different tables are found: 
- [ ] /nettleie/alt-om-nettleie/fleksibelt-forbruk
- [ ] /mitt-elvia/prisoversikt
- [ ] /minside/kundeforhold/7473c97e-40ee-4341-8dac-6489b1bf6b38/malere


packages/ui/src/components/Table/Table.module.scss
```diff
@@ -8,26 +8,40 @@ $table-border-radius: 8px;
 %header-style {
   background: map-get($colors, 'white');
   width: 100%;
-  height: spaceCalc(7);
   th {
     color: map-get($colors, 'black');
     font-weight: bold;
     font-size: emCalc(18px);
+    height: spaceCalc(7);
```
steinmy, line 15: This is because setting the height of thead doesn't work in Chrome. Not related to the other fixes.

steinmy, line 15: This is because setting the height of thead doesn't work in Chrome. Not related to the other fixes.

packages/ui/src/components/Table/Table.module.scss
```diff
@@ -8,26 +8,40 @@ $table-border-radius: 8px;
...
+}
+
+.limited-width-table-container {
+  @include standardWidth;
+
+  margin: spaceCalc(3) auto;
+}
+
 .e-table {
-  display: table;
```
steinmy, line 20: `display: table;` was not needed, as it is the default anyways

packages/ui/src/components/Table/Table.tsx
```diff
@@ -49,24 +51,32 @@ export const Table: FC<TableProps> = ({
...
-        className
-      )}
+    <div
+      className={classnames(css.outerTableContainer, {
+        [css.limitedWidthTableContainer]: isLimitedWidth,
+      })}
     >
-      {children}
-    </table>
+      <div className={css.innerTableContainer}>
```
steinmy, line 59: The diff here is confusing, but the only actual change is adding the two divs

packages/web/src/templates/Meterpoints/MeteringPoints.tsx
```diff
@@ -23,29 +23,27 @@ export const MeteringPoints: React.FC<MeteringPointsProps> = ({
   )
 
   return (
-    <div className={css.tableContainer}>
```
steinmy, line 26: The only change here is removing the div. It is no longer needed, as table containers are now built in to the component.

steinmy (commented):
Added clarifying comments for reviewers

lobunto:
Seems like the small table variant on Min side has got som borders... Screenshot 2020-09-16 at 15 42 44

eTallang:
The code in itself LGTM. Let's just check that it doesn't introduce unintended side effects in the existing tables :slightly_smiling_face:

steinmy:
> Seems like the small table variant on Min side has got some surrounding borders... 
Oh no! I'll look into it.

steinmy:
> 
> 
> > Seems like the small table variant on Min side has got some surrounding borders...
> 
> Oh no! I'll look into it.

Should be fixed now

lobunto (approved):

PR #97, opened 2020-09-17
Fix/omega 169 redirect issue min side

lobunto:
OMEGA-169

Customers may experience that they are being redirected to the access request page even if they have 1 or more customer relationships.

Users with 0 customer relationships should be redirected to the access request page.

## How is it solved?

Redirecting the user from the `Customer.ts` store before it is passed to the CustomerRealtionOverview component.

- [ ] Users with 0 customer relationships should be redirected to the access request page
- [ ] The CustomerRelationOverview should not render content before it know if we actually know that we some customer relationships.


packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.tsx
```diff
@@ -14,6 +14,8 @@ export const CustomerRelationOverview: React.FC = () => {
 
   const allCustomerRelations = useSelector(customersSelectors.getAllCustomers)
   const isLoading = useSelector(customersSelectors.getIsLoading)
+  const customerIds = Object.keys(allCustomerRelations)
+  const showCustomerRelationView = !isLoading && customerIds.length > 1
 
   Object.keys(allCustomerRelations).forEach(key => {
```
ragnhildlarsen, line 20: Since `Object.keys(allCustomerRelations)` are stored in a const now, maybe it can be used for the forEach loop as well?
`customerIds.forEach(key => {`

lobunto: Good spot! Sure we can do that :)

lobunto: Fixed @ragnhildlarsen

t0ff3r (approved):

PR #98, opened 2020-09-17
fix(ui): seems to fix the missing module error for ui

t0ff3r:

imports from '@elvia/ui' gives errors about module not found and auto complete does not work.

Should work

## How is it solved?

Added another index.ts on root in ui package

- [x] Autocomplete and imports in web from ui should work without vscode complaining


erlbe (approved):
The fix seems to work for me. No error on module import and autocomplete seems to work. Tested in Form.tsx

lobunto:
@t0ff3r will you fix the conflicting files?

PR #100, opened 2020-09-17
Fix/omega 94 checkbox validation bug

ragnhildlarsen:
OMEGA-94

In checkbox list at least one is required and should not show error.

How to reproduce:
- Go to https://www.elvia.no/proff/verktoy-og-skjema/skjemaer/tilknytning-av-storre-anlegg-eller-flytting-av-stromnettet
- Check both checkboxes
- Uncheck one checkbox and see error message

image-20200903-073253

Another bug that was found during investigation for this task was that error message was shown below each checkbox:
Screenshot 2020-09-17 at 11 34 17

Checkbox list should not show error message if at least one checkbox is checked. 

Screenshot 2020-09-17 at 11 29 57

Only one error message should be shown below the group of checkboxes.

## How is it solved?
- Added a check in `isRequiredAndDirty` in `FormInputUtils.tsx` that checks if one of the checkboxes are checked. Error is only shown if none of the checkboxes are checked. 
- To ensure that only one error message is shown, I added another condition in the check in `hideError` to not be shown if options are more than one in `InputOptionGroup.tsx`.  

- [ ] Error is shown if no checkboxes are checked and they are touched
- [ ] Error is not shown if at least one checkbox is checked in a group


packages/ui/src/components/FormInputs/FormInputUtils.tsx
```diff
@@ -28,11 +28,19 @@ export const useRequiredAndDirty = <T,>(props: InputProps<T>) => {
...
+    )
+  }
+
   if (value && !isDirty) {
     setIsDirty(true)
   }
 
-  const showRequired = isRequired && isEmpty && (isDirty || !!props.showError)
+  const showRequired =
+    isRequired && isEmpty && (isDirty || !!props.showError) && !isOneChecked
```
t0ff3r, line 43: Is `!!props.showError` necessary? showError seems to be a boolean according to `InputProps`

t0ff3r: So you could maybe just write  `props.showError`?

ragnhildlarsen: I agree :)

t0ff3r (approved):

PR #101, opened 2020-09-18
feat(ui): add range to DatePicker

lobunto:
OMEGA-112

PR 1 of 2 for OMEGA-112 "add date range for graph".
The DatePicker component should later be moved out to a separate package. It should probably also be used as a hook.

By providing a property `showRange` you can choose a range between two dates.
The result returned from the change event is an object containing `from` and `to` date.

## How is it solved?

Extend the DatePicker component to also return an object with dates instead of just a string.
Add separate styles config for normal DatePicker and DatePicker with range.

- [ ] You should be able to choose a range between two dates
- [ ] Both selected dates should appear in the input box
- [ ] The design should be according to the design sketches in [figma](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=10567%3A72840) (design QA has been done before the PR was created).


packages/ui/src/components/Datepicker/Datepicker.tsx
```diff
@@ -158,16 +180,59 @@ export const Datepicker: React.FC<DatepickerProps> = props => {
...
+          to: rangeState.to || rangeState.from,
+        },
+      ]
+    } else {
+      if (isValidDate) {
+        return dayjs(value, format).toDate()
+      } else {
+        return null
+      }
+    }
```
t0ff3r, line 233: Should be possible to flatten these nested if/else to simplify a bit?

packages/ui/src/components/Datepicker/Datepicker.tsx
```diff
@@ -158,16 +180,59 @@ export const Datepicker: React.FC<DatepickerProps> = props => {
...
+      callOnRangeChange(range)
+      setRangeState(range)
+    }
+
+    if (!showRange && !(modifiers.disabled || modifiers.outside)) {
+      callOnChange(dayjs(day).format(format))
+
+      setTimeout(() => {
+        setShowDatepicker(false)
+      }, 100)
```
t0ff3r, line 214: Is the timeout necessary?

t0ff3r (approved):
Code looks good except for a few minor comments that are not show stoppers but in general I find this component a too big and complex. We should break out some hooks and clean this one up a bit. Not saying you need to fix it right now. Good work!

lobunto:
> Code looks good except for a few minor comments that are not show stoppers but in general I find this component a too big and complex. We should break out some hooks and clean this one up a bit. Not saying you need to fix it right now. Good work!

Good! Will make a package of it in a separate PR.

packages/ui/src/components/Datepicker/Datepicker.tsx
lobunto: I cannot see what function it does and there is no difference by removing it.

lobunto: Well, the datepicker is actually a bit laggy without the timeout when clicking on a date. So added it back for now.

PR #102, opened 2020-09-18
Feature/pilot dashboard block component

JonatanPaalsson:
OMEGA-126

Pilot customers should get a dashboardblock with information regarding the new graph and communication strategy

Pilotusers should get a new dashboardblock right below the pilotgraph

## How is it solved?

This is solved with a new PilotDashboard component. I wanted to use a separate component for Pilots in order to easy structure core connected to the pilot. Here we choose the correct terms from contentful depending on your tariffgroup and communication strategy

image

- [ ] Users without pilot should not be shown the new dashboardblock
- [ ] Users with pilot should get a dashboardblock with information connected to your tariffgroup and communication strategy


packages/web/src/components/PilotDashboardBlock/PilotDashboardBlock.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { DashboardBlock, Typography, TypographyType } from '@elvia/ui'
+import { getTerm, Terms, getRichTerm } from '@utils/getTerm'
+import * as css from './PilotDashboardBlock.module.scss'
+
+interface PilotDashboardBlockProps {
+  tariffGroup: number
+  communicationStrategy: number
```
lobunto, line 8: Should we probably have some types for what numbers these can have, instead of just `number`?

lobunto (changes requested):

lobunto (approved):

packages/web/src/components/PilotDashboardBlock/PilotDashboardBlock.module.scss
```diff
@@ -0,0 +1,3 @@
+.textBlock {
+  overflow-wrap: anywhere;
```
eTallang, line 2: It doesn't look like this is supported in IE11. Whats our strategy/goal for IE support?

eTallang: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

eTallang: Probably not a big deal though, since this is just an initial PR for the new features?

packages/web/src/components/PilotDashboardBlock/PilotDashboardBlock.tsx
```diff
@@ -0,0 +1,43 @@
...
+    getRichTerm(Terms.minSidePilotRushAndCalmEnvironmentText),
+    getRichTerm(Terms.minSidePilotRushAndCalmCostText),
+  ]
+
+  const title = allTitles[props.tariffGroup - 1]
+  const body =
+    allBodys[props.tariffGroup * (props.communicationStrategy + 1) - 1]
+
+  return (
+    <div>
```
eTallang, line 36: This div can probably be removed

packages/web/src/components/PilotDashboardBlock/PilotDashboardBlock.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <DashboardBlock title={blockTitle}>
+        <Typography type={TypographyType.TITLE_SMALL}>{title}</Typography>
```
eTallang, line 38: I think this component can be replaced by the `TitleSmall`-component for better readability.

eTallang (changes requested):

packages/web/src/components/PilotDashboardBlock/PilotDashboardBlock.module.scss
JonatanPaalsson: Nice catch!
Our strategy is that it should functionally work but we do close to nothing of testing or maintaining. For example the current frontpage dosen't load any images in IE11

eTallang: So lets keep it then :slightly_smiling_face:

eTallang (approved):
LGTM

PR #103, opened 2020-09-18
OMEGA-168: Adjust metering point page to new design

eTallang:
OMEGA-168

Previously, the customer name and navigational tabs were visible on the metering point page, which is confusing.
image

This PR changes this by hiding the title, customer no, and tabs when on the metering point page. It also adds the selected metering point to the breadcrumbs for navigational purposes.

[Url or screenshot if possible]
image

- [ ] Check that the implementation complies with the design in [Figma](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=11089%3A77531).
- [ ] Check that the breadcrumbs update when selecting other metering points, and when navigating back and forth.
- [ ] Check that the breadcrumbs can be used for navigation.
- [ ] Check that the page is responsive


lobunto (dismissed):

t0ff3r:
If you only have 1 meterpoint you should get it in the tab for "Min måler" and not get the table. Right?

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -50,6 +57,14 @@ export const CustomerDetail: React.FC<CustomerDetail> = ({
...
   ]
 
+  const getContractAddressWithId = (id: string) => {
+    const contract = contracts.find(c => c.id === id)
+    const address = contract?.meteringPoint.address
+    return `${address?.streetName} ${
+      address?.buildingNumber
+    }${address?.letter || ''}`
+  }
+
```
marasva, line 67: Should / could this be written in extractDataHelper.tsx or something?

lobunto: There is actually a Ramda function in extractDataHelper.tsx that does exactly this: `getCustomerAdress()`

eTallang:
> If you only have 1 meterpoint you should get it in the tab for "Min måler" and not get the table. Right?

True, but this will be solved through OMEGA-147

t0ff3r (approved):

PR #107, opened 2020-09-21
Feature/omega 31 new access overview

ragnhildlarsen:
OMEGA-31

Accesses to contracts should be shown on a customer relation level, instead of on contract/meterpoint level as it currently is today.

[Figma sketches](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=1%3A33)

## How is it solved?

A new tab with "Tilganger" should be visible on a customer detail page. 

Added a new folder in template for the new view called `AccessOverview`. I used the Table component for web to get the correct structure of the table on mobile. The table should have more columns if it is a private customer relationship, than if it is an organisational/"Bedrift". The column for last logged in date is not supported by the backend yet. 

Because of the new data structure in the backend, I needed to change the structure of how the requests are stored in the state. 

The information box only includes the link to requesting a private customer relationship because requesting access to an organisation customer relation is not supported yet. 

**The new data structure for access request is only available in dev environment so this PR needs to be tested locally.**

- [ ] Petter Pettersen(birth number 07060570129) have access to Per Hansen(birth number 05039703930)
- [ ] Per Hansen should be able to revoke Petter Pettersen's access
- [ ] Petter Pettersen should not be able to revoke Per Hansen's access or his own access
- [ ] Access overview looks identical with the design defined in Figma
- [ ] Table shows correct accesses to the customer relationship
- [ ] Roles is displayed correct for private customer relationships
- [ ] Revoke access request is only available for owners


packages/web/src/templates/AccessOverview/AccessOverview.tsx
```diff
@@ -0,0 +1,329 @@
...
+  const contractsAndOwners: ContractsAndOwner[] = []
+
+  let isCustomerOwnerOfContract = false
+
+  contracts.forEach(contract => {
+    const contractsAndOwner = useSelector((state: RootState) =>
+      contractsSelectors.getReceivedContractsAndOwner(state, contract.id)
+    )
+    contractsAndOwners.push(contractsAndOwner)
+    isCustomerOwnerOfContract = useSelector((state: RootState) =>
```
JonatanPaalsson, line 111: Won't this just set `isCustomerOwnerOfContract` for the last contract in the list?

packages/web/src/templates/AccessOverview/AccessOverview.tsx
```diff
@@ -0,0 +1,329 @@
...
+      customersSelectors.isCustomerOwnerOfContract(state, contract.id)
+    )
+  })
+
+  const customerAccessList: CustomerAccess[] = []
+
+  const isOwnerAlreadyInList = (id: string): boolean => {
+    return customerAccessList.findIndex(
```
JonatanPaalsson, line 119: Maybe use some instread of findIndex? Something similar to this...
`return customerAcessList.some(customerAccess => customerAccess.id === id)`

packages/web/src/templates/AccessOverview/AccessOverview.tsx
```diff
@@ -0,0 +1,329 @@
...
+      {
+        value: <p>{terms.personsWithAccessHeader}</p>,
+      },
+      {
+        value: <p>{terms.phoneHeader}</p>,
+      },
+    ]
+  }
+
+  const tableData = [tableHeader]
```
JonatanPaalsson, line 231: isn't tableHeader already an array?

packages/web/src/templates/AccessOverview/AccessOverview.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  useEffect(() => {
+    if (accessToken) {
+      dispatch(fetchAccessRequestReceived(accessToken))
+    }
+  }, [accessToken])
+
+  let tableHeader
+  if (customerType === CustomerTypes.PERSON) {
+    tableHeader = [
```
JonatanPaalsson, line 206: Can't this just be an array of HTMLElements?

JonatanPaalsson (changes requested):
There seems to be a lot of functions in `AccessOverview` that try to select specific data from redux. Do you think some of these could be selectors instead?

packages/web/src/store/accessRequestSent/accessRequestSent.ts
```diff
@@ -58,10 +51,7 @@ const accessRequestSent = createSlice({
       const accessRequest = action.payload
       state.isRequesting = false
       state.isSubmitted = true
-      state.sentRequestsById[accessRequest.id] = accessRequest
-      state.sentRequestsAllIds = union(state.sentRequestsAllIds, [
-        accessRequest.id,
-      ])
+      state.sentRequests[0].accessRequests.push(accessRequest)
```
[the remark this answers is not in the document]
ragnhildlarsen: sentRequests is the requests sent by the user logged in. Since the new data structure aggregates all requests from a user into one object this should only contain one accessRequestWrapper. I'm not sure that doing this state change is necessary at all, because sentRequests are retrieved from the backend when a user enters minside, so it might be possible to remove it completely.

packages/web/src/templates/AccessOverview/AccessOverview.tsx
ragnhildlarsen: Tried to remove `{ value: }` and just list the `<p>` tags, but the Table component expect it to be at the format that it is now.

packages/web/src/templates/AccessOverview/AccessOverview.tsx
ragnhildlarsen: Moved this outside the forEach loop so it should be more correct now.

packages/web/src/templates/AccessOverview/AccessOverview.tsx
```diff
@@ -0,0 +1,312 @@
...
+    role: RoleTypes.OWNER,
+  }
+  customerAccessList.push(customerAccess)
+
+  const selectedCustomerContractIds = contracts.map(contract => contract.id)
+  const filterRequestsBasedOnSelectedCustomerIds = (
+    requestsToBeFiltered: AccessRequestWrapper[]
+  ) => {
+    return requestsToBeFiltered.filter(req => {
+      const contractIds = req.accessRequests.map(r => r.contractIds)
```
lobunto, line 120: `req.accessRequests` seems to be undefined since `accessRequests` is no prop on the requests sent into this function

lobunto (changes requested):

packages/web/src/templates/AccessOverview/AccessOverview.tsx
ragnhildlarsen: This is because the backend changes are only in the dev environment so it does not work to test via the netlify link.

lobunto: Ok, it seemed to look fine when I tested it yesterday.

packages/web/src/templates/AccessOverview/AccessOverview.tsx
```diff
@@ -0,0 +1,312 @@
...
+    ),
+    accessToken = useSelector((state: RootState) =>
+      oidcSelectors.getAccessToken(state)
+    ),
+    requestsSent = useSelector((state: RootState) =>
+      accessRequestSentSelectors.getSentRequests(state)
+    ),
+    requestsReceived = useSelector((state: RootState) =>
+      accessRequestReceivedSelectors.getReceivedRequests(state)
+    )
```
lobunto, line 103: Can we have the const keyword before each variable here? It's hard to see what uses selectors and not. All the variables using the `useSelector` could also be moved into a single `useSelector`

Something like this:
```
const {
    requestsSent,
    requestsReceived,
    accessToken,
    contracts,
    ownerId,
    customer,
  } = useSelector((state: RootState) => ({
    customer: customersSelectors.getCustomerById(state, customerId),
    ownerId: customersSelectors.getOwnerId(state),
    contracts: contractsSelectors.getContractsForCustomer(state, customerId),
    accessToken: oidcSelectors.getAccessToken(state),
    requestsSent: accessRequestSentSelectors.getSentRequests(state),
    requestsReceived: accessRequestReceivedSelectors.getReceivedRequests(state),
  }))
```

lobunto (changes requested):

packages/web/src/templates/AccessOverview/AccessOverview.tsx
lobunto: Probably we can make a function that maps over table elements and wraps them with the `value-object`

lobunto: Thinking something like this could work: 
```
const tableHeaders = [terms.personsWithAccessHeader, terms.phoneHeader]
 
const tableValueObjects = tableHeaders.map(term => ({ value: <p>{term}</p> }))

const tableData = [tableValueObjects]
```

packages/web/src/components/Table/Table.tsx
```diff
@@ -55,6 +55,7 @@ function restructureTableData(children: any) {
 }
 
 export const Table: FC<TableProps> = ({ data, className, isLimitedWidth }) => {
+  console.log(data)
```
JonatanPaalsson, line 58: Ops

JonatanPaalsson (approved):

lobunto (approved):

PR #110, opened 2020-09-21
Include mobileNumberSender in sms access request

ragnhildlarsen:
Related to OMEGA-31

Backend supports receiving mobileNumberSender when a user sends an access request, but the frontend only sends the mobileNumber of the user receiving the request.

The new access overview table created in task OMEGA-31 shows phone number of the users that have access so the owner have a way to contact the people that have received access.

[Figma sketches](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=10157%3A1825)

Requesting access by sms should work as before. The response from backend should include mobileNumberSender.

## How is it solved?

Retrieved the info of the customer that is the owner, because this is the person that is logged in. Included the mobilePhone value in the action dispatched for requestAccessBySms.

It might be a good idea to go through the complete flow of requesting and granting access by sms to make sure that adding the mobileNumberSender does not break anything.


corneliusgdahling (approved):
Looks good

PR #112, opened 2020-09-22
Feature/omega 163 poweroutage map legend

erlbe:
OMEGA-163

Display and explain the colours of the poweroutage map.

Displaying a legend which shows what the orange and red colour means.

Desired result displayed in Figma: https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=51%3A1

## How is it solved?

Add a Legend component to Poweroutage package. Uses termContext to get the terms to display.
image

- [ ] /strombruddskartv2 should display the legend. However not placed on the map yet.


packages/poweroutage/src/components/Legend/Legend.module.scss
```diff
@@ -0,0 +1,42 @@
+@import '../../../../ui/src/theme';
```
t0ff3r, line 1: This should probably be imported in some other way?

packages/poweroutage/src/index.tsx
```diff
@@ -108,6 +109,9 @@ export const PowerOutage: React.FC<PowerOutageProps> = ({ terms }) => (
           {table}
         </InfoBox>
       </div>
+      <div style={{ width: '336px' }}>
```
t0ff3r, line 112: Should this be in a class instead?

packages/poweroutage/src/components/Legend/Legend.module.scss
FredricW, line 1:
```suggestion
@import '@elvia/ui/src/_theme.scss';

Funker =)
```

erlbe: Da bruker jeg det 👍

packages/poweroutage/src/index.tsx
erlbe: Added css styling in class to set width and position the legend.

t0ff3r (approved):

FredricW (approved):

PR #115, opened 2020-09-24
fix(web): changed property that starts video, so works on initial load build mobile

corneliusgdahling:
OMEGA-189

This property is better, also, because it does not start the video, before it *thinks* it does not need to re-buffer. This however makes the video load a bit slower. Especially on slow internet connection and large file sizes (obviously).

I got so caught up in trying to make the previous property `onLoadedData` to work (I still have no idea why it did not work, it should have), that I did not consider there might be another way, a better way.


JonatanPaalsson:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadeddata_event) this might not be fired on mobile devices if the browser has enabled data-saver. Could that be the original issue?

JonatanPaalsson (changes requested):
According to [MDN ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadeddata_event) this might not be fired on mobile devices if the browser has enabled data-saver. Could that be the original issue?

If that is the case maybe we shouldn't load the video?

corneliusgdahling:
> According to [MDN ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadeddata_event) this might not be fired on mobile devices if the browser has enabled data-saver. Could that be the original issue?
> 
> If that is the case maybe we shouldn't load the video?

I don't have any of those things enabled on any devices, I tested. But I guess it is possible, that is the reason.

JonatanPaalsson (approved):

PR #116, opened 2020-09-24
Feature/make datepicker package

lobunto:

`Datepicker` component has become large and should be moved to a separate package.

`Datepicker` component and it's caption component are moved to the `datepicker` package

## How is it solved?

Moved `datepicker` component and it's `caption` component to the `datepicker` package
Added `package.json` file for `datepicker` package
Added `datepicker` package as a dependency for `ui` package

- [ ] No changes in prod
- [ ] Test the datepicker in a form (eg. search for "Bestill nettleie for bedrift")
- [ ] Datepicker component are moved to a separate package `datepicker`


packages/datepicker/src/index.tsx
```diff
@@ -1,33 +1,41 @@
...
 import classnames from 'classnames'
-import DayPicker from 'react-day-picker'
-import { LocaleUtils } from 'react-day-picker'
-import 'react-day-picker/lib/style.css'
-import dayjs from 'dayjs'
-import 'dayjs/locale/nb'
+import css from './index.module.scss'
 import customParseFormat from 'dayjs/plugin/customParseFormat'
-import { Input, InputError, InputLabel, FormInputProps } from '../Input'
+import dayjs from 'dayjs'
```
t0ff3r, line 8: This should probably be a dependency in this package.

packages/datepicker/src/index.tsx
```diff
@@ -1,33 +1,41 @@
+import 'dayjs/locale/nb'
+import 'react-day-picker/lib/style.css'
+import DayPicker from 'react-day-picker'
```
t0ff3r, line 3: This should probably be a dependency in this package.

packages/datepicker/src/index.tsx
```diff
@@ -1,33 +1,41 @@
...
-import dayjs from 'dayjs'
-import 'dayjs/locale/nb'
+import css from './index.module.scss'
 import customParseFormat from 'dayjs/plugin/customParseFormat'
-import { Input, InputError, InputLabel, FormInputProps } from '../Input'
+import dayjs from 'dayjs'
+import themeStyles from './index.theme.scss'
+import { DateUtils } from 'react-day-picker'
 import { HddCustomCloseMenu } from '@hafslundnett/hdd-icons/hdd-icons-react'
+import { LocaleUtils } from 'react-day-picker'
```
t0ff3r, line 12: Could maybe merge a few of these imports? Think this works? `import DayPicker, { LocaleUtils } from 'react-day-picker'`

packages/datepicker/src/index.tsx
```diff
@@ -208,10 +216,7 @@ export const Datepicker: FC<DatepickerProps> = props => {
 
     if (!showRange && !(modifiers.disabled || modifiers.outside)) {
       callOnChange(dayjs(day).format(format))
-
-      setTimeout(() => {
```
t0ff3r, line 212: Nice

t0ff3r (changes requested):
Looks great! Nice to get this complex thing out of ui. Make sure to clean up dependencies here so we manage those in this package and not in ui and its good. :-)

packages/datepicker/src/index.tsx
lobunto: Yep

lobunto:
@t0ff3r everything is fixed :)

t0ff3r:
@lobunto Reason this is not merged is that there are some concerns regarding how we decide if something should be a package. Grab a hold of me and @FredricW for a discussion regading this :-)

PR #120, opened 2020-09-24
fix(contentful-source): adds transformer for features

t0ff3r:
OMEGA-206

Contentful-source transformer needs to fetch features from contentful

...

## How is it solved?

Adds a new field transformer and fetcher for features


packages/contentful-source/gatsby-node.js
```diff
@@ -144,6 +144,19 @@ exports.createSchemaCustomization = ({ actions, schema }) => {
         infer: true,
       },
     }),
+    schema.buildObjectType({
+      name: 'ElviaFeature',
+      fields: {
+        name: 'String!',
+        isVisible: 'Boolean!',
+        id: 'ID!',
+        typeName: 'String',
```
FredricW, line 153:
```suggestion
        typeName: 'String!',
```
Denne er vel "non-nullable" også?

FredricW (approved):

PR #122, opened 2020-09-25
feat(web): add history label in MeteringPointRow

lobunto:
OMEGA-221

Add history label for contracts with an `endTime` 

In the metering point list, contracts with an `endTime` should get an "Historisk måler" label displayed instead of monthly consumption.

Screenshot 2020-09-25 at 11 22 58

## How is it solved?

Checks if a contract has `endTime` not `null`

- [ ] The "Historisk måler" label is visible for contracts with `endTime`
- [ ] The current monthly consumption (or "-") is visible for contracts without `endTime`


packages/web/src/templates/Meterpoints/MeteringPointRow.tsx
```diff
@@ -62,6 +63,13 @@ export const MeteringPointRow: React.FC<MeterpointRowProps> = ({
     return currentConsumption && `${Math.round(currentConsumption)} kWh`
   }
 
+  const hasEndTime = complement(propEq('endTime', null))
+  const isHistory = ifElse(
+    hasEndTime,
+    always(<Label data-testid="Label">Historisk måler</Label>),
```
eTallang, line 69: What's the purpose of the `data-testid`-attribute?

packages/web/src/templates/Meterpoints/MeteringPointRow.tsx
```diff
@@ -62,6 +63,13 @@ export const MeteringPointRow: React.FC<MeterpointRowProps> = ({
     return currentConsumption && `${Math.round(currentConsumption)} kWh`
   }
 
+  const hasEndTime = complement(propEq('endTime', null))
```
eTallang, line 66: I'm not sure what benefits rambda provides here over a regular `if/else`-expression

eTallang (changes requested):

packages/web/src/templates/Meterpoints/MeteringPointRow.tsx
```diff
@@ -62,6 +63,13 @@ export const MeteringPointRow: React.FC<MeterpointRowProps> = ({
...
+  const isHistory = ifElse(
```
steinmy, line 67: The `isHistory` variable serves two purposes: indicating whether the meterpoint is expired, and containing the label. How about instead building a `consumptionFieldContent` which is set to either the label or `getCurrentMonthsConsumption()` depending on whether `hasEndTime` is true or false?

steinmy (dismissed):

packages/web/src/templates/Meterpoints/MeteringPointRow.tsx
lobunto: `data-testid` is supposed to make it easier for selecting stuff when testing, it is not special for React in any way. But I actually will remove it here since it is already added to the Label component in `ui`

packages/web/src/templates/Meterpoints/MeteringPointRow.tsx
```diff
@@ -11,8 +11,9 @@ import {
 import { RootState } from '@store/root-reducer'
 import { getUrlWithParams, routes } from '@routing'
 import { consumptionsSelectors, fetchConsumptions } from '@store/consumptions'
-import { useHover, useMedia, MediaTypes, ElvisIcon } from '@elvia/ui'
+import { useHover, useMedia, MediaTypes, ElvisIcon, Label } from '@elvia/ui'
 import css from './Meterpoints.module.scss'
+import { always, complement, ifElse, propEq } from 'ramda'
```
eTallang, line 16: I believe these are unused now

eTallang (changes requested):

packages/web/src/templates/Meterpoints/MeteringPointRow.tsx
lobunto: yep, fixed :)

eTallang (approved):
LGTM

packages/web/src/templates/Meterpoints/MeteringPointRow.tsx
```diff
@@ -63,6 +63,12 @@ export const MeteringPointRow: React.FC<MeterpointRowProps> = ({
     return currentConsumption && `${Math.round(currentConsumption)} kWh`
   }
 
+  const consumptionFieldContent = contract.endTime ? (
+    <Label>Historisk måler</Label>
+  ) : (
+    getCurrentMonthsConsumption() || '-'
+  )
+
```
t0ff3r, line 71: As `getCurrentConsumption` already returns a string I don`t see why that function should not handle returning '-' aswell?

lobunto: Sure, just pushed a commit with fixes. Take a look at it @t0ff3r

PR #125, opened 2020-09-25
Chore/omega 100 clean up routing

eTallang:
OMEGA-100

Previously, routing on min side was done through a combination of entries in `basePaths.json` and hard coded strings.

This change removes all hard coded strings and introduces a single source for all routes.

## How is it solved?

A new file `routing.ts` is introduced that contains all routes and a few utility functions to create and manipulate urls.


packages/web/src/routing.ts
```diff
@@ -0,0 +1,79 @@
...
+  const urlParts = url.split('/')
+
+  if (!urlParts.length) {
+    return ''
+  }
+
+  const lastPart = urlParts.pop()!
+
+  if (lastPart.startsWith(':')) {
+    if (url === routes.meteringPoint);
```
steinmy, line 51: What is happening here? It seems like this line does nothing, or perhaps I'm misunderstanding?

steinmy (changes requested):

packages/web/src/routing.ts
t0ff3r: I agree

t0ff3r (approved):
Think this looks great! Maybe remove that if doing nothing and consider if /faktura should be /fakturaer for consistency in single/plural. Other than that I'd say ship it :-)

packages/web/src/routing.ts
```diff
@@ -0,0 +1,79 @@
...
+  faq: 'sporsmal-og-svar',
+  myPage: '/minside',
+  customerPage: `/minside/:${customerIdToken}`,
+  meteringPoints: `/minside/:${customerIdToken}/malere`,
+  meteringPoint: `/minside/:${customerIdToken}/malere/:${contractIdToken}`,
+  customerInfo: `/minside/:${customerIdToken}/kundeinfo`,
+  accessOverview: `/minside/:${customerIdToken}/tilganger`,
+  hanPort: `/minside/:${customerIdToken}/hanport/:${contractIdToken}`,
+  hanPortSuccess: `/minside/:${customerIdToken}/hanport/success`,
+  invoices: `/minside/:${customerIdToken}/faktura/:${contractIdToken}`,
```
t0ff3r, line 35: /faktura could maybe be fakturaer since we use tilganger/malere

packages/web/src/routing.ts
eTallang: Good catch. Strange that it ended up in the PR. Removed.

marasva:
If I try to log in with Randi Rystad or Ali Jensen in test and my own account in prod I get this error message: 
`Error: Number of provided params to getUrlWithParams does not match with required url params.`

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.tsx
```diff
@@ -27,7 +28,7 @@ export const CustomerRelationOverview: React.FC = () => {
 
   useEffect(() => {
     if (!isLoading && customerIds.length === 1) {
-      navigate(`/minside/kundeforhold/${customerIds[0]}`)
+      navigate(getUrlWithParams(routes.myPage, customerIds[0]))
```
marasva, line 31: In relation to my previous comment it worked for me by just changing routes.myPage to routes.customerPage here.

eTallang: A very good catch.

PR #126, opened 2020-09-25
fix(web): changes default language to norwegian

t0ff3r:
OMEGA-233

Screenreaders reads wrong language

They should not

## How is it solved?

Changes default lang to norwegian

- [ ] Check lang prop in html-tag


eTallang (approved):
LGTM

PR #127, opened 2020-09-25
OMEGA-227: Get access requests related to customer id

ragnhildlarsen:
OMEGA-227

The Access Overview table currently only shows access requests related to the person logged in (request it has sent or request it has received). If a user have access to a customer relationship/contract it should see all the other people that have access to the same customer relationship.

Petter Pettersen should be able to see all other people that also have access to Per Hansens customer relationship, like Randi Rystad. Same for Randi, she should see Petter in the list when she looks at Per Hansen's customer relationship. Per Hansen, who is the owner of the customer relationship should see both Petter and Randi in the table.

Screenshot 2020-09-25 at 15 46 02

## How is it solved?

This is solved by using the new endpoint `/accessrequest/customer/{customerId}` instead of `/me/accessrequest/received` that was previous used. The requests are fetched with the customerId from the url when the user clicks the Tilganger tab.

**The accesses described here is testable in the dev environment.**

- [ ] Petter Pettersen should be able to see all other people that also have access to Per Hansens customer relationship, like Randi Rystad.  
- [ ] Same for Randi, she should see Petter in the list when she looks at Per Hansen's customer relationship.
- [ ] Per Hansen, who is the owner of the customer relationship should see both Petter and Randi in the table.


packages/web/src/store/accessRequestReceived/accessRequestReceived.ts
```diff
@@ -61,12 +63,16 @@ const accessRequestReceived = createSlice({
 })
 
 export const fetchAccessRequestReceived = (
+  customerId: string,
```
eTallang, line 66: This method should probably be renamed. E.g. to `fetchAccessRequestsByCustomerId`

packages/web/src/store/accessRequestReceived/accessRequestReceived.ts
```diff
@@ -61,12 +63,16 @@ const accessRequestReceived = createSlice({
...
   accessToken: string
 ): AppThunk => async dispatch => {
   const { actions } = accessRequestReceived
   try {
     dispatch(actions.getAccessRequestReceivedStart())
```
eTallang, line 71: This action should probably be renamed. E.g. to `getAccessRequestsByCustomerIdStart`. Same with the success- and failure-actions.

eTallang: Hmm, or is the `ByCustomerId`-part implementation-details? Would it be possible to rename them simply `getAccessRequestsStart()` and so forth?

eTallang (changes requested):

packages/web/src/store/accessRequestReceived/accessRequestReceived.ts
eTallang: Following the logic of the comment below, maybe we could rename this `fetchAccessRequests`?

packages/web/src/store/accessRequestReceived/accessRequestReceived.ts
ragnhildlarsen: renamed to `accessRequests`

packages/web/src/store/accessRequestReceived/accessRequestReceived.ts
ragnhildlarsen: renamed to `accessRequests`

eTallang (approved):
LGTM

PR #128, opened 2020-09-28
Feature/OMEGA-220 Add metering point description to metering points table

ragnhildlarsen:
OMEGA-220

Metering points table should have a column for metering point description according to the [Figma sketches](https://www.figma.com/file/sJHoScmQRq0hlr6AcHI9WB/Elvia.no?node-id=9560%3A55962 - Connect to preview)

Metering point description should be shown if the metering point description has a value, otherwise it should show `-`

Screenshot 2020-09-28 at 11 06 50

## How is it solved?

Solved by adding a new term in Contentful for "Målerbeskrivelse" that was used by a new `<th>` tag. In the MeteringPointRow a new `<td>` tag was added that accessed the metering point description value. 

- [ ] Metering point description should be shown in the table
- [ ] If the metering point description value is null, `-` should be shown


ragnhildlarsen:
Had to merge PR for OMEGA-146 to include some changes related to using terms in the table.

eTallang:
You probably want to convert this to a draft, since it depends on the filter task being merged :slightly_smiling_face:

t0ff3r (approved):

t0ff3r:
Does the description exist for any meters yet? Or is this column empty for everyone?

packages/web/src/services/customer-portal-api.ts
```diff
@@ -105,9 +105,9 @@ export interface MeteringPoint {
   meterSerialNumber: string
   meteringPointTypeValue: string
   isAms: boolean
+  meteringPointDescription: string
```
eTallang, line 108: A description is probably optional (should be confirmed)? Should in that case be marked with `?`.

eTallang (changes requested):

ragnhildlarsen:
> Does the description exist for any meters yet? Or is this column empty for everyone?

I guess some meters may have a description but the only way to set a description is to contact customer service so probably not many meters that have it.

packages/web/src/services/customer-portal-api.ts
ragnhildlarsen: When testing the endpoint in swagger it seems like it always returns contracts with metering points that have the field meteringPointDescription. The value is null if description is not set.

eTallang: Maybe we could set the type to `string | null` in that case then :slightly_smiling_face:

t0ff3r (approved):

eTallang (approved):
LGTM

PR #129, opened 2020-09-28
rename Tab component to Segment

marasva:

Tab component is not actually tabs and is thus renamed. The renaming is done due to creation of a new Tab component.

There should not be any changes that are viewable.

### How is it solved?
Check everything looks the same.


packages/ui/src/components/Segment/Segment.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [firstTab, setFirstTab] = useState(true)
+  return (
+    <SegmentGroup>
+      <Segment active={firstTab} onClick={() => setFirstTab(true)}>
+        First Segment
+      </Segment>
+      <Segment active={!firstTab} onClick={() => setFirstTab(false)}>
+        Second Segment
+      </Segment>
+    </SegmentGroup>
```
t0ff3r, line 17: Maybe rename state here aswell?

packages/web/src/components/Header/PublicHeader.tsx
```diff
@@ -396,7 +396,7 @@ export const PublicHeader: React.FC<PublicHeaderProps> = ({
     return (
       <>
         <div className={css.mainPageMobileMenu}>
-          <TabGroup>{customerSegmentTabs}</TabGroup>
+          <SegmentGroup>{customerSegmentTabs}</SegmentGroup>
```
t0ff3r, line 399: maybe just customerSegments if it does not affect too much code?

t0ff3r (commented):
Looks good. Maybe fix the minor comments and we're good?

packages/web/src/components/Header/PublicHeader.tsx
marasva: yes, fixed

t0ff3r (approved):

PR #130, opened 2020-09-28
add cypress for interface testing

JonatanPaalsson:
OMEGA-211

To make the project a bit more stable and to remove manual testing on the website, we have decided to implement some automated tests. For this Cypress was the framework that we have decided to use. 

Any developer should now be able to run cypress locally.

## How is it solved?

By adding cypress as a new package

- [ ] Run `yarn cypress:open` while having a local version of the project running
- [ ] Run and verify all new tests


packages/cypress/cypress/integration/frontpage/shortcuts.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { isNot404Page } from '../../utils'
+
+context('All shortcuts on frontpage should work', () => {
+  beforeEach(() => {
+    cy.visit('')
+  })
+
+  it('Navigate to each shortcut from the frontpage and then back', () => {
+    cy.get('.SectionShortcuts-module--shortcut--1bsiR').each(($el, index) => {
```
eTallang, line 9: We should probably use `[data-cy]`-selectors instead of class selectors (referencing the [Best Practices](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements). Should we in that case introduce some general rule for adding `[data-cy]`-attributes in the HTML upfront, or should it be added when writing the Cypress tests?

packages/cypress/cypress/plugins/index.ts
```diff
@@ -0,0 +1,21 @@
+/// <reference types="cypress" />
```
eTallang, line 1: Can this file be removed?

packages/web/src/components/Search/SearchField.tsx
```diff
@@ -53,6 +53,7 @@ export const SearchField: React.FC<SearchFieldProps> = ({
       <input
         className={css.input}
         data-testid="search-input"
```
eTallang, line 55: What are the `data-testid`-attributes for?

eTallang (changes requested):

packages/cypress/cypress/integration/frontpage/search.spec.ts
```diff
@@ -0,0 +1,15 @@
+context('Search', () => {
+  beforeEach(() => {
+    cy.visit('')
+  })
+
+  it('Navigate to first result in search on desktop', () => {
+    cy.wait(500)
```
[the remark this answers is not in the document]
JonatanPaalsson: Yes, the thing is that our website is not intractable for some time once it has loaded. So even if it finds the element in the DOM, it might not be able to click on it. When removing the wait the test fails about 50% of the time 🤷

packages/cypress/cypress/plugins/index.ts
JonatanPaalsson: It can. Same for `commands` and `fixtures`. These are some files that were provided on install that i decided to keep in order to keep the suggested folder-structure and to show some potential features of cypress. What do you think?

packages/web/src/components/Search/SearchField.tsx
JonatanPaalsson: Not sure actually, might be for old puppeteer tests or some tracking 🤷

packages/cypress/cypress/plugins/index.ts
eTallang: We had great use of the `commands` at my previous project. Commands are really useful to create custom commands that is added to the global `cy`-object. For example, we added a `login`-method that allowed us to do the following:
```
cy.login('admin@trim.com').<additional chainable calls>
```

Another good example was the `goToPage`-method:
```
cy.goToPage('Customers').get('[data-cy]="title"')...
```
It's a really handy way of adding general-purpose actions that is likely to be repeated in many tests, while it also ensures that  the next step is not performed before the previous is finished.

I'm pretty sure we'll get some commands in this project sooner or later, so I vote for keeping the `commands` at least :smiley:

packages/web/src/components/Search/SearchField.tsx
eTallang: I see them all around, while I'm pretty sure we don't have tests that use them. It's not related to this PR though, but I suggest that we have a chat about them in the team, and decide whether to keep them and use them, or remove them.

packages/cypress/cypress/integration/frontpage/search.spec.ts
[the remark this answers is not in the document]
eTallang: Hmm that's weird. I feel like we should try to solve this Cypress-challenge now, before we start writing lots of tests. I have bad experience with introducing `wait`s in tests, where the tests in that project after a while took over 40 mins to run, since the tests had `wait`s all over.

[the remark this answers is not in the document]
JonatanPaalsson: Definitely agree. I removed most of the waits, but I can't remove this one at this time. Our current header is rerendering 2 times on desktop because of the `useMedia`-hook used in the header. This hook assumes that the device is on a mobile (since most of our users are on mobile), and this causes a rerender when we "switch" to desktop.

So what happens in the test is that cypress finds the element and tries to click it. But if there is a rerender between these two operations, then the test fails. Resolving the rerender issue would be to rewrite a lot in the header which I don't want to do in this pr, so therefor I keept the wait. If you have any other ideas how this could be solved, I'd love to hear it :)

eTallang (approved):
LGTM

packages/web/src/components/Search/SearchField.tsx
FredricW: They exists for unit testing.

t0ff3r (approved):

PR #133, opened 2020-09-29
Feature/omega 147 redirect to metering page

marasva:
[OMEGA-147](https://elvia.atlassian.net/secure/RapidBoard.jspa?rapidView=17&projectKey=OMEGA&modal=detail&selectedIssue=OMEGA-147)

For private customers who mainly have only one meterpoint it is confusing to get it in a table first. Therefore customers with only one meterpoint is redirected straight to the meterpoint site. 

image

 
image

## How is it solved?

Change CustomerRelationOverview to redirect only if the customer have one customerRelations and one contract (Randi Rystad). Added changes to CustomerCard if the customer has several customerRelations but only one meterpoint on each card the customer should be redirected to the meterpoint page straight away (Per Hansen's access to Randi Rystad). 
Changes to how the tabs behave since they do not get rerouted to the correct address since the hierarchy is a bit changed when some customers are skipping a link. 
Change a bit so the breadcrumbs behave correctly.

Have used the new implementation of tabs, so it is set as draft until the tabs are in production. 

- [ ] Check that is has the expected behavior for all the test users (Randi Rystad, Ali Jensen, Per Hansen)
- [ ] Check that is has the expected behavior for your own user in prod
- [ ] Check breadcrumbs


steinmy (changes requested):
Looks good to me, but I found some refactoring opportunities

packages/web/src/components/CustomerCard/CustomerCard.tsx
```diff
@@ -10,20 +10,46 @@ import {
...
+export const CustomerCard: React.FC<CustomerCardProps> = ({
+  customer,
+  contracts,
+}) => {
   const isDesktop = useMedia(MediaTypes.DESKTOP)
 
+  const customerContracts =
+    contracts !== null
+      ? contracts.filter(c => c.customerId === customer.id)
+      : null
```
steinmy, line 30:
```suggestion
  const customerContracts = contracts?.filter(c => c.customerId === customer.id)
```
Here the optional chaining operator `?.` can be used to achieve the same effect more conciesly

packages/web/src/components/CustomerCard/CustomerCard.tsx
```diff
@@ -10,20 +10,46 @@ import {
...
+
+  const onlyHaveOneContract =
+    customerContracts !== null &&
+    customerContracts !== undefined &&
+    customerContracts.length === 1
+      ? true
+      : false
```
steinmy, line 37:
```suggestion
  const onlyHaveOneContract = customerContracts?.length === 1
```

packages/web/src/components/CustomerCard/CustomerCard.tsx
```diff
@@ -10,20 +10,46 @@ import {
...
+  const contractId =
+    onlyHaveOneContract &&
+    customerContracts != null &&
+    customerContracts[0].id !== null
+      ? customerContracts[0].id
+      : null
```
steinmy, line 43:
```suggestion
  const contractId =
    onlyHaveOneContract 
      ? customerContracts[0]?.id
      : null
```

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -81,7 +88,7 @@ export const CustomerDetail: React.FC<CustomerDetailProps> = ({
         url: uri,
       })
 
-      if (meteringPointPageMatch) {
+      if (meteringPointPageMatch && !onlyHaveOneContract) {
```
steinmy, line 91: I know this has nothing to do with your changes, but while trying to understand the code, it took a while for me to realize what `meteringPointPageMatch` means. If I'm not wrong, it indicates whether the current page shown is the metering point page. Could it perhaps be renamed `currentPageIsMeteringPointPage`. In the same vein, maybe `onlyHaveOneContract` could be renamed `customerOnlyHasOneContract`. Then this line would read "If current page is metering point page and not customer only has one contract" which would explain quite clearly in close to natural language what is actually happening.

steinmy (changes requested):
Some more refactoring suggestions

packages/ui/src/components/Tabs/Tabs.tsx
```diff
@@ -96,7 +105,11 @@ export const Tabs: React.FC<TabsProps> = ({ tabs = [], uri }) => {
         >
           {tabs?.map((tab, index) => (
             <Link
-              to={uri + '/' + tab.url}
+              to={
+                customerOnlyHasOneContract && tab.url === meteringPointsRoute
+                  ? `${uri}/${tab.url}/${contractId}`
+                  : uri + '/' + tab.url
```
t0ff3r, line 111: Didn't @eTallang create a helper for creating urls?

packages/ui/src/components/Tabs/Tabs.tsx
```diff
@@ -107,7 +120,11 @@ export const Tabs: React.FC<TabsProps> = ({ tabs = [], uri }) => {
                 }
               }}
             >
-              <TextCaps className={css.tabContent}>{tab.name}</TextCaps>
+              <TextCaps className={css.tabContent}>
+                {customerOnlyHasOneContract && tab.name === 'Målere'
+                  ? 'Måler'
+                  : tab.name}
```
t0ff3r, line 126: Tab names should be controlled by the cms

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.tsx
```diff
@@ -28,7 +36,15 @@ export const CustomerRelationOverview: React.FC = () => {
...
-      navigate(getUrlWithParams(routes.customerPage, customerIds[0]))
+      contracts.length === 1
+        ? navigate(
+            getUrlWithParams(
+              routes.meteringPoint,
+              customerIds[0],
+              contracts[0].id
+            )
+          )
+        : navigate(getUrlWithParams(routes.customerPage, customerIds[0]))
```
t0ff3r, line 47: I think probably the router should handle what page you end up on, and not be nested navigations in children components. We should discuss this with @eTallang

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.tsx
```diff
@@ -56,7 +76,11 @@ export const CustomerRelationOverview: React.FC = () => {
           </TextCaps>
           <div className={css.sectionWrapper}>
             {companyCustomerRelations.map(relation => (
-              <CustomerCard key={relation.id} customer={relation} />
+              <CustomerCard
+                key={relation.id}
+                customer={relation}
+                contracts={contracts}
```
t0ff3r, line 82: Maybe CustomerCard needs to select this data from the store instead of having it passed down everywhere?

t0ff3r (changes requested):

packages/ui/src/components/Tabs/Tabs.tsx
marasva: Yes, have talked with Steinar about that.

packages/web/src/templates/CustomerRelationOverview/CustomerRelationOverview.tsx
marasva: Customer is passed down from before, but removed the contracts being passed down. Should the customer also be removed?

t0ff3r (approved):

PR #137, opened 2020-09-30
feat(ui): set +47 as default country code for phone inputs

eTallang:
OMEGA-153

Currently, there is no default for country code in phone inputs.
image

This change introduces +47 as a default value for country code in phone inputs.
image

- [ ] Check that the default value is +47
- [ ] Check that the value can be changed
- [ ] Check that the new value that is entered is registered in the form data


t0ff3r (approved):
Nice work!

PR #138, opened 2020-09-30
Feature/omega 192 han port tabs

steinmy:
OMEGA-X

There are tabs in the new design. The HAN page didn't use them until now.

image
image
image

HAN page shows tabs, and there is a tab for going to the HAN page
image

Back link is gone from the bottom of the HAN page
image

Section with link to HAN page is removed from meter point page, and meter point details list is moved to where this box was before. If the meter is an old meter, box for inputing readings is shown below invoice table.
image

The HAN tab is only shown if the user has at least one active meter with a HAN port (AMS).

## How is it solved?

I don't really understand this question. It is solved by making the things shown above...

- [ ] Log in to the customer portal and see that things are as described above
  - Use Ali Jensen for a customer with HAN ports
  - Use Randi Rystad for a customer with no HAN ports


packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -45,19 +51,26 @@ export const CustomerDetail: React.FC<CustomerDetailProps> = ({
...
       url: getRelativeUrl(routes.profile),
     },
     {
-      name: 'Tilganger',
+      name: getTerm(Terms.minSideTabsAccess),
       url: getRelativeUrl(routes.accessOverview),
     },
   ]
 
+  if (activeContractsWithHanPort.length > 0) {
```
JonatanPaalsson, line 67: Does this mean that we don't show the HAN-tab for inactive contracts?

packages/web/src/templates/Meterpoint/Meterpoint.module.scss
```diff
@@ -45,17 +45,10 @@
   }
 }
 
-.columns {
-  display: grid;
+.manualReadingBlock {
+  max-width: 364px;
```
JonatanPaalsson, line 49: use emCalc :)

JonatanPaalsson (changes requested):

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
steinmy: Yes, the HAN-port list only shows active contracts, and so if you have no active contracts, then there are no elements in the list, and thus no reason to show the tab.

packages/web/src/templates/Meterpoint/Meterpoint.module.scss
steinmy: Oh, I thought emCalc was only used for font size, but I must have misunderstood. Will fix.

t0ff3r:
@steinmy is this design approved?

steinmy:
> 
> 
> @steinmy is this design approved?

No, it is disapproved, so I'll have to fix it. The invoices list becomes too narrow now that the meterpoint page has a changed layout.

JonatanPaalsson (changes requested):
If i log in with Per Hansen, I get to see the Han-tab for every customer that Per Hansen has access to even if that customer has no meterpoints with HAN. Is this intended? It's a bit hard to validate when the tab should show and not..

steinmy:
@JonatanPaalsson 
> If i log in with Per Hansen, I get to see the Han-tab for every customer that Per Hansen has access to even if that customer has no meterpoints with HAN. Is this intended? It's a bit hard to validate when the tab should show and not..

This is a known issue, and will be fixed by https://elvia.atlassian.net/browse/OMEGA-250 and is thus not in the scope of this pull request. Randi Rystad can be used as an example of a user for whom the HAN tab should not be shown.

JonatanPaalsson (approved):

packages/web/src/components/MeterPointInfo/MeterPointInfo.module.scss
```diff
@@ -9,3 +9,11 @@
     fill: map-get($colors, 'black');
   }
 }
+
+.header {
+  font-weight: 500;
```
FredricW, line 14: Is there no typography that matches the text?

steinmy: I'm not sure. Earlier <b> was used, but it set the font-weight to 700 instead of 500. I can do some research on this.

packages/web/src/components/MeterPointInfo/MeterPointInfo.tsx
```diff
@@ -16,113 +16,118 @@ import {
...
+  customerId: string
 }
 
-export const MeterPointInfo = ({ contract }: MeterpointInfoProps) => {
-  const { hanPortEvent, loading, contractById } = useSelector(
+interface TableRowData {
+  headerKey: string
+  headerContent: string | ReactElement
+  fieldKey: string
+  fieldContent: string | ReactElement
```
FredricW, line 30:
```suggestion
  headerContent: React.ReactNode
  fieldKey: string
  fieldContent: React.ReactNode
```
This type includes everything that can be rendered by react and is the parent type for the ones you have used.

packages/web/src/components/MeterPointInfo/MeterPointInfo.tsx
```diff
@@ -16,113 +16,118 @@ import {
...
+          getTerm(Terms.minSideHanPort)
+        ),
+      fieldKey: 'hanPortStatus',
+      fieldContent: hanEventIsLoading
+        ? 'Laster...'
+        : getHANStatusText(hanPortEvent?.status),
       id: '3',
     })
 
+  const meterpointInfoTableRows = meterpointInfoTableData.map(
```
FredricW, line 102: I like the idea of standardizing the creation of table rows, but I think you should take it even further and make this a proper function and return the whole table row template in its final form. Hit me up on Discord if it is unclear.

packages/web/src/store/hanPortEvent/selectors.ts
```diff
@@ -14,28 +14,31 @@ export const getHanPortEventsByAllContracts = (state: RootState) =>
...
-    .map(contract => ({
-      contract: contract,
-      hanPortEvent: hanPortEvents.find(
-        event => event?.contractId === contract.id
-      ),
-    }))
+  const mergedContractsAndHanPortEvents = contracts.map(contract => ({
+    contract: contract,
+    hanPortEvent: hanPortEvents.find(
+      event => event?.contractId === contract.id
```
FredricW, line 39: Can an event in the list, actually be `null` or `undefined`?

packages/web/src/components/MeterPointInfo/MeterPointInfo.tsx
```diff
@@ -16,113 +16,118 @@ import {
...
-            <b>{getTerm(Terms.minSideHanPort)}</b>
-          </TextInfo>
-        </span>,
-        <TextInfo key="hanPortStatus" tag={TypographyTag.SPAN}>
-          {getHANStatusText(hanPortEvent?.status)}
-        </TextInfo>,
-      ],
+      headerKey: getTerm(Terms.minSideHanPort),
+      headerContent:
+        contractById.endTime === null ? (
```
FredricW, line 85: what does it mean when `endTime` is `null`? That the contract has ended?

packages/web/src/components/MeterPointInfo/MeterPointInfo.tsx
FredricW: That means that I want it to return the final jsx so that you don't need the additional map functions in template as well

packages/web/src/components/MeterPointInfo/MeterPointInfo.tsx
steinmy: Oh, thanks!

PR #141, opened 2020-10-01
feat(web): scroll to error when han port submission fails

eTallang:
OMEGA-182

Today, if you make an error in the HAN port form (e.g. not checking any checkboxes) and press "Bestill åpning", an error box appears on the top of the form. However, this is hard to see for the user, since its outside of the viewport.

image

This change scrolls the user to the error box, so that it is visible for the user.


eTallang:
> Have you checked if this applies to other forms, and the fix should be implemented there as well?

Yes, this should only apply to this form. Other forms are created through contentful, and all of those scroll to the top on error automatically.

JonatanPaalsson (approved):

PR #143, opened 2020-10-01
fix(web): incorrect height between article blocks

eTallang:
OMEGA-164

Currently, it is hard to quickly get a sense of paragraphs in articles, due to a small line height between paragraphs.
image

This PR fixes this issue by increasing the spacing between paragraphs from 8px to 24px.
image


JonatanPaalsson:
Should this be for Category and MainCategory aswell?

JonatanPaalsson (changes requested):
Should this be for Category and MainCategory aswell? Or maybe even more..?

eTallang:
> Should this be for Category and MainCategory aswell? Or maybe even more..?

Mabe? Haven't seen any other places where the layout breaks, however, as you say, I see that it's specified 8px margin top on paragraphs in the Category-component as well. I don't have enough knowledge about the repo yet to know every place this might be an issue. Should we fix it where we know about the issue, and create new issues whenever we find new occurences?

JonatanPaalsson:
Maybe check with the creator of the issue? Is there some way to validate this for expandableBlock that the Jira-issue specifies?

eTallang:
I see that the increased line height makes for example this article look strange:
https://5f75cd4d8f111042f60f717f--elvia-test.netlify.app/drift-og-vedlikehold/strombrudd/strombrudd-dette-er-de-vanligste-arsakene

It looks just as strange in prod today, just more condense. The reason for the large spacing between a title and the text below, is that each title is defined as its own paragraph. This issue would be solved if the title for each paragraph was within the `p`-element. Is this something that should be discussed with the content team?

JonatanPaalsson:
Not really sure what is looking strange, but this is probably because I'm blind as a fucking bat. I'm marking this as waiting for design approval since I think it's their call if this looks good enough or not. I'm not a real fan of putting a `h3` inside a `p`, but we can of course discuss this based on input from design 😄

eTallang:
> Not really sure what is looking strange, but this is probably because I'm blind as a fucking bat. I'm marking this as waiting for design approval since I think it's their call if this looks good enough or not. I'm not a real fan of putting a `h3` inside a `p`, but we can of course discuss this based on input from design

Hehe, I was thinking of this part:
image

It's hard to easily get an understanding of the groups of text there.

JonatanPaalsson (approved):

t0ff3r (approved):

PR #146, opened 2020-10-02
Feature/omega 187 show polygons if poweroutage is focused

corneliusgdahling:
OMEGA-187

**NB: Using `Record<string, any>` for map and maps from google maps api, because adding `@types/googlemaps` was not a trivial task. Created task OMEGA-249 to fix types**

Show Polygon marking poweroutage, if focused.

Show poweroutage polygon if focused.

## How is it solved?

Uses google maps api to create polygon and add to map

[Example screenshot of the solution if possible]
image

Click on a chip and see if polygon shows (along with info box).
If closing info box, chip should show and polygon should hide.


t0ff3r:
Ready to merge is maybe not the correct label when it is not reviewed yet? 😄

packages/poweroutage/src/index.tsx
```diff
@@ -46,31 +47,41 @@ export const PowerOutage: React.FC<PowerOutageProps> = ({ terms }) => {
...
-                  <Pin
-                    type={powerOutage.type}
-                    affectedCustomers={powerOutage.affectedCustomers}
-                    onClick={() => setSelected(powerOutage)}
-                  />
-                </Marker>
-              ))}
+          <Map
+            defaultBounds={bounds}
+            polygons={selected ? [selected.polygon] : undefined}
```
FredricW, line 62:
```suggestion
            polygons={selected && [selected.polygon]}
```

FredricW (approved):
Fix that minor check and I think this is good! Great work :)

PR #148, opened 2020-10-02
feat(web): makes it possible to implement feature toggles

t0ff3r:
OMEGA-206

We need a way to use feature toggles in web

A new component for wrapping features should be available for developers

## How is it solved?

Adds a component called EnabledFeature to web, and adds 3 examples of usage to dev page

- [ ] Check out dev page in test and compare to enabled features in contentful (test)


packages/web/src/components/EnabledFeature/EnabledFeature.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { useFeatures } from '@queries/useFeatures'
+
+interface EnabledFeatureProps {
+  name: string
+  enabledComponent?: React.ReactNode
```
JonatanPaalsson, line 6: You can change these to `React.ReactElement` instead, then you don't need the fragment in the code later. Doesn't seem to work for children tho 🤷

JonatanPaalsson (changes requested):

JonatanPaalsson (approved):

PR #151, opened 2020-10-05
refactor(poweroutage): moved polygon logic to hook and added types

FredricW:
OMEGA-X

GoogleMap component got a little crowded and there is some issues with the current polygon logic that makes it not always trigger or clean out properly.

Polygons should now also work when clicking on an item in the poweroutages list. The polygon should always disappear when leaving the details view.

## How is it solved?

Moved the logic to a new `usePolygons` hook. It takes the google maps api instances and return a `setPolygons`- and a `removePolygons` function.
Also imported google.maps types and applied them into our `GoogleMap` component.

- [x] Clicking on a pin should show a polygon
- [x] Closing the details view should hide the polygon
- [x] Clicking on a item in the poweroutages list should show a polygon


packages/ui/src/components/GoogleMap/GoogleMap.tsx
```diff
@@ -12,10 +13,11 @@ export interface Coordinate {
 
 export type Boundary = NESWBounds
 
-export type GoogleMap = Record<string, any>
-export type GoogleMaps = Record<string, any>
+export type GoogleMap = google.maps.Map
+// eslint-disable-next-line no-undef
+export type GoogleMaps = typeof google.maps
```
FredricW, line 18: This one is a bit weird. It complains that `google` is not defined, yet it exists as a global namespace. Using the `GoogleMaps` type later on does indeed yield the correct types so idk what the fuzz is all about 8)

corneliusgdahling (approved):

t0ff3r (approved):

erlbe (approved):

PR #154, opened 2020-10-06
feat(web): redesign login page

eTallang:
OMEGA-222

Currently, there is multiple different sign in methods on the sign in page, depending on what kind of user you are.
image

This PR changes this by being the first step of a new sign in experience. The new design is hidden behind a feature flag, and should not be enabled before the full sign-in flow has been implemented.

image


packages/web/src/templates/MainCategory/MainCategory.tsx
```diff
@@ -101,7 +99,10 @@ export const MainCategory: React.FC<PageQueryResponse> = ({
           ))}
         </Breadcrumbs>
         <div className={css.mainCategoryHeader}>
-          <PageTitle title={title} />
+          <PageTitle
+            title={title}
+            coloredDots={pageContext.currentPath === '/logg-inn'}
```
t0ff3r, line 104: Don't think we should tie a path here as this can be changed if the category in cms is renamed

eTallang: I agree, so I actually took the liberty of always displaying colored dots. Will have to consult this with design.

eTallang: The path should have been referenced from the routing object anyways :sunglasses:

lobunto (approved):
LGTM

t0ff3r:
Looks good, but verify that it works by enabling in test before merge 😄

t0ff3r (approved):

PR #157, opened 2020-10-07
ci: builds contentful extensions with CI and deploys to Netlify

t0ff3r:

Contentful extensions needs package upgrades to improve security and also we need to build and deploy it to netlify due to size restrictions on contentful hosted extensions

Less security notifications and easier maintainance of extensions from now on

## How is it solved?

Adds a new pipeline config for extensions and deploys to two new pages in netlify.

- [x] Verify that table still works in contentful test by making a table, publish and verify
- [x] Verify that use-form-fields still works in contentful test by making a form, publish and verify


CI/azure-pipeline.extensions.yml
```diff
@@ -0,0 +1,99 @@
+trigger:
+  branches:
+    include:
+      - trunk
+  paths:
+    include:
+    - extensions/*
```
erlbe, line 7: Since we are using branch-naming conventions to trigger different pipelines. Maybe they should be documented in the readme file? Like `extensions/*` and `migrations/*`

t0ff3r: This is path triggers, not branch naming :-) Only care about triggering the build if a changed happened in this folder.

erlbe: Ah, yes! Never mind then 😄

erlbe (approved):

CI/azure-pipeline.extensions.yml
```diff
@@ -0,0 +1,99 @@
...
+            displayName: 'Extract BuildTable zip'
+            inputs:
+              archiveFilePatterns: '$(Pipeline.Workspace)/BuildTable/**.zip'
+              destinationFolder: $(System.DefaultWorkingDirectory)/build/table
+          - task: ExtractFiles@1
+            displayName: 'Extract BuildUseFormFields zip'
+            inputs:
+              archiveFilePatterns: '$(Pipeline.Workspace)/BuildUseFormFields/**.zip'
+              destinationFolder: $(System.DefaultWorkingDirectory)/build/use-form-fields
+          - script: 'node scripts/netlify.js deploy -s $(siteId) -a $NETLIFY_TOKEN -d build -m "#$(Build.BuildId) - $(Build.SourceBranchName)" --prod'
```
JonatanPaalsson, line 96: In the regular pipeline we use `$(prodFlag)`, should we use the same here? Not really sure about the difference 🤷

JonatanPaalsson (changes requested):

CI/azure-pipeline.extensions.yml
t0ff3r: We add --prod on all builds here as it would not make sense to have deploy previews for this since we do not have lots of environments in contentful.

CI/azure-pipeline.extensions.yml
```diff
@@ -0,0 +1,99 @@
...
+    jobs:
+      - job: BuildArtifactTable
+        displayName: 'Build table extension'
+        pool:
+          vmImage: 'ubuntu-latest'
+        variables:
+          - group: elvia-extensions
+        steps:
+          - script: |
+              cd extensions/table
```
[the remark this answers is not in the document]
t0ff3r: Thanks! Fixed :-)

JonatanPaalsson (approved):
Nice work!

PR #159, opened 2020-10-08
test(inputlocation): check that maps shows in inputlocation

JonatanPaalsson:

Yestersday we resolved an issue where inputlocation did not show a map. So here is a test to validate this

Reduce regression

## How is it solved?

By adding a test that checks "tre-over-linje"-form for an inputlocation, clicks it and expects a map


t0ff3r (approved):

corneliusgdahling (approved):
Good work! :D

PR #163, opened 2020-10-08
feat(customerdetail): add back button on customer relation and meter point page

ragnhildlarsen:
OMEGA-231 (& OMEGA-235)

Should be easier to navigate back from customer relation pages to customer relation overview and from meter point page back to meter points table.

Customer relation pages and meterpoint page should have a back button that navigates the user up one level.
Screenshot 2020-10-08 at 11 24 30
Screenshot 2020-10-08 at 11 24 11

## How is it solved?

This was initially two separate tasks, but after discussing with @eTallang we decided that solving both cases in `CustomerDetail.tsx` would be a good approach.
The Link component from the ui package was used to create the back button. A new term was added in Contentful to separate from other back buttons on my page. 
The link is shown only when the customer relation has multiple meter point. 

- [ ] Back button is shown on customer relation page when there are multiple customer relationship
- [ ] Back button is shown on meter point page when there are multiple meter points on the customer relationship
- [ ] Back button should not be shown when there is just one customer relationship with one meter point


packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -132,7 +140,12 @@ export const CustomerDetail: React.FC<CustomerDetailProps> = ({
           <Tabs tabs={tabs} />
         </>
       )}
-      {children}
+      <div className={css.contentContainer}>{children}</div>
+      {!customerOnlyHasOneContract && (
+        <ElviaLink href="../" type={LinkType.Back}>
+          {backButton}
+        </ElviaLink>
+      )}
```
steinmy: Wouldn't it be simpler to just put a class on the link with a margin-top? Then there would be no need for an extra div.

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -132,7 +140,15 @@ export const CustomerDetail: React.FC<CustomerDetailProps> = ({
           <Tabs tabs={tabs} />
         </>
       )}
-      {children}
+      <div className={css.contentContainer}>{children}</div>
+      {(!customerOnlyHasOneContract || Object.keys(customers).length > 1) && (
```
steinmy, line 144: `Object.keys(customers).length > 1` happens on line 91 as well. Maybe it could be deined as a const, just like `customerOnlyHasOneContract`

packages/web/src/templates/CustomerDetail/CustomerDetail.module.scss
```diff
@@ -24,3 +24,7 @@
 .header {
   margin: 0 spaceCalc(3) 0 0;
 }
+
+.contentContainer {
+  margin-bottom: spaceCalc(8);
+}
```
steinmy, line 30: Maybe the link could have margin-top instead of the content having a margin-bottom? Then there would be no need for the extra div around the content.

steinmy (changes requested):
Seems nice, but I have a few suggestions

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
ragnhildlarsen: I agree :)

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
```diff
@@ -45,9 +51,12 @@ export const CustomerDetail: React.FC<CustomerDetailProps> = ({
   const customerNoLabel = `${getTerm(Terms.minSideProfileCustomerNoText)} ${
     customer?.cabCustomerNumber
   }`
+  const backButton = getTerm(Terms.minSideCustomerDetailBackButton)
+
   const location = useLocation()
 
   const customerOnlyHasOneContract = customerContracts?.length === 1
+  const moreThanOneCustomer = Object.keys(customers).length > 1
```
steinmy, line 59: Nå er jeg veldig pirkete her, men hva med `hasAccessToMoreThanOneCustomerRelation`?

steinmy (changes requested):

packages/web/src/templates/CustomerDetail/CustomerDetail.tsx
ragnhildlarsen: Enig, var ikke helt fornøyd med det navnet jeg heller 😄

PR #165, opened 2020-10-08
refactor(googlemap and geolocationmap): Use native google maps api to set zoom and center

erlbe:
OMEGA-282

Change the way zoom and centering in the map works. This gives us more control over the interactions in the map.

Map to work as normal :)

## How is it solved?

Use the native Google maps api to set zoom and center (via fitBounds) instead of google-map-react, which left the developers confused.

- [x] Verify that GeoLocationMap works (/drift-og-vedlikehold/meld-fra-om-farlige-forhold/tre-over-linje)


packages/ui/src/components/GeoLocationMap/GeoLocationMap.tsx
```diff
@@ -191,24 +191,11 @@ export const GeoLocationMap: React.FC<GeoLocationMapProps> = ({
...
-        </div>
-      )}
-    </div>
-  )
-
   return (
     <div data-testid="GeoLocationMap" className={classes} style={style}>
       <GoogleMap
         center={center ? center : undefined}
+        autoFitBoundsToContent={false}
```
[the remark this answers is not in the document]
erlbe: Done

corneliusgdahling (approved):
LGTM lets :shipit: ... nvm. need more reviewers ⏳

packages/ui/src/components/GoogleMap/GoogleMap.tsx
```diff
@@ -49,25 +62,33 @@ export const GoogleMap: React.FC<GoogleMapProps> = ({
...
 
   useEffect(() => {
-    if (polygons) {
+    if (!googleMap || !googleMaps || !autoFitBoundsToContent) return
+    if (markers.length || polygons.length) {
+      const markerCoordinates: Coordinate[] = markers.map(m => m.coordinate)
+      const polygonCoordinates: Coordinate[] = polygons.reduce(
+        (prev, next) => [...prev, ...next],
+        []
+      ) // flatten
```
t0ff3r, line 84:
```js
const polygonCoordinates = flatten(polygons) // Ramda ftw!
```

packages/ui/src/components/GoogleMap/GoogleMap.tsx
```diff
@@ -49,25 +62,33 @@ export const GoogleMap: React.FC<GoogleMapProps> = ({
...
-    : {
-        center: defaultCenter,
-        zoom: defaultZoom,
-      }
 
   useEffect(() => {
-    if (polygons) {
+    if (!googleMap || !googleMaps || !autoFitBoundsToContent) return
+    if (markers.length || polygons.length) {
+      const markerCoordinates: Coordinate[] = markers.map(m => m.coordinate)
```
t0ff3r, line 80:
```js
const markerCoordinates: Coordinate[] = markers.map(prop('coordinate')) // Ramda FTW!
```

t0ff3r (approved):
Would prefer some more ramda love when it makes sense but other than that i looks good.

packages/ui/src/components/GoogleMap/GoogleMap.tsx
erlbe: Agreed!

packages/ui/src/components/GoogleMap/GoogleMap.tsx
erlbe: Agreed!

PR #167, opened 2020-10-08
feat(pilot): add enableFeature

JonatanPaalsson:
OMEGA-269

We should be able to control when we want to release the pilotproject

Pilot should not be seen in prod, only in test.
In test the pilotgraph should also be shown

## How is it solved?

With the enabledFeature component.

- [ ] Should show only in test


t0ff3r (approved):
Nice

corneliusgdahling (approved):

